PR #2, opened 2026-02-09
Feat/railgun gas benchmark

JohnGuilding:
## What does this PR do:
* Scaffold a foundry project 
* Add initial benchmark for Railgun using a real tx as test data
* Record snapshots and saves them to `RailgunSendBenchmark.json`
* Add CI to run fmt, build and tests

## How to test this PR
1. Run `forge test -vv`
2. Run `cast receipt 0x9691b06aeb81ae709ad478942a5307c5e5cc4597074f5b60e11b2033c1584580 gasUsed --rpc-url $ETH_RPC_URL` to observe gas cost is similar to real tx
3. Observe gas cost is similar to real tx 

## Future work
* Add proof generation logic so we can use a locally generated proof
* Add proof generation benchmarks logic

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,8 +4,7 @@ permissions: {}
 
 on:
   push:
-  push:
-      branches: [main]
+    branches: [main]
   pull_request:
   workflow_dispatch:
 
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2026 PSE
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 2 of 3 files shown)


.gitmodules
NicoSerranoP: Should we pin a specific version/commit/tag for the external packages? What will happen if they change their contract and we are not notified?

snapshots/RailgunSendBenchmark.json
NicoSerranoP: Add a line at the end of the file. I was able to configure this to happen automatically in VS Code by going to:

Settings -> Files -> Insert Final Newline -> enabled

NicoSerranoP (approved):
Looks good! I just left a couple of non-blocking comments to consider 🚀

test/RailgunTransact.t.sol
```diff
@@ -0,0 +1,135 @@
...
+        vm.startSnapshotGas("railgun_send");
+        RELAY.transact(txs);
+        uint256 gasUsed = vm.stopSnapshotGas("railgun_send");
+
+        console.log("scenario:  railgun_send");
+        console.log("forkBlock:", block.number);
+        console.log("gasUsed:  ", gasUsed);
+    }
+
+    function _buildTransaction() internal pure returns (Transaction memory) {
```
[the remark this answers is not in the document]
JohnGuilding: If following the solidity style guide, leading underscore [signifies a private or internal function](https://docs.soliditylang.org/en/latest/style-guide.html#underscore-prefix-for-non-external-functions-and-variables)

.gitmodules
JohnGuilding: Agreed, pinned to latest commit

snapshots/RailgunSendBenchmark.json
JohnGuilding: I think this can be resolved in a future formatting PR where we add formatting and linting rules. Added an issue https://github.com/privacy-ethereum/private-transfers-benchmarks/issues/3

PR #17, opened 2026-02-18
feat: add lowDB + human-readable events + git hooks

NicoSerranoP:
- Use lowDB (noSQL simple database) to have concurrent/fast benchmarking
- Run protocol benchmarks in parallel via Promise.all
- Each protocol's benchmark() now returns metrics instead of writing to file;
  results are saved once at the end via db.write()
- Convert all ABI event definitions in constants.ts files to more human-readable format
- Added husky hooks for git commits to prevent lint or format errors

With the help of https://claude.ai/code/session_01KXqn1R93nH6g145aq29khg

### DIFF
--- a/gas-benchmarks/pnpm-workspace.yaml
+++ b/gas-benchmarks/pnpm-workspace.yaml
@@ -1,2 +1,5 @@
 ignoredBuiltDependencies:
   - unrs-resolver
+
+onlyBuiltDependencies:
+  - esbuild
--- a/gas-benchmarks/src/index.ts
+++ b/gas-benchmarks/src/index.ts
@@ -1,16 +1,42 @@
 import { PrivacyPools } from "./privacy-pools/index.js";
 import { Railgun } from "./railgun/index.js";
 import { TornadoCash } from "./tornado-cash/index.js";
+import { db } from "./utils/db.js";
 
 const railgun = new Railgun();
 const tornadoCash = new TornadoCash();
 const privacyPools = new PrivacyPools();
 
+await db.read();
+
 const start = Date.now();
 
-await railgun.benchmark();
-await tornadoCash.benchmark();
-await privacyPools.benchmark();
+const [railgunMetrics, tornadoCashMetrics, privacyPoolsMetrics] = await Promise.all([
+  railgun.benchmark(),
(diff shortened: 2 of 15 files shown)


JohnGuilding (approved):
Looks great!

gas-benchmarks/pnpm-workspace.yaml
```diff
@@ -1,2 +1,5 @@
 ignoredBuiltDependencies:
   - unrs-resolver
+
+onlyBuiltDependencies:
```
JohnGuilding: What does this do?

gas-benchmarks/src/index.ts
```diff
@@ -1,16 +1,17 @@
...
 const privacyPools = new PrivacyPools();
 
+await db.read();
+
 const start = Date.now();
 
-await railgun.benchmark();
-await tornadoCash.benchmark();
-await privacyPools.benchmark();
+await Promise.all([railgun.benchmark(), tornadoCash.benchmark(), privacyPools.benchmark()]);
```
JohnGuilding: We're running `db.update` concurrently, is that safe to do? Could there be a race condition if the rpc requests complete at the same time?

I don't know how lowdb works or if it protects against this. If it is a risk, a simple change could be to return the values from each method call and call .update once in `index.ts`

gas-benchmarks/pnpm-workspace.yaml
```diff
@@ -1,2 +1,5 @@
 ignoredBuiltDependencies:
```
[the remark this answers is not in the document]
NicoSerranoP: It indicates pnpm which packages to build and which ones to ignore. The build process is required for packages that have different libraries for each OS (e.g. Windows, Linux). This file is automatically generated after running `
pnpm approve-builds` which usually appears when installing a new dependency

gas-benchmarks/pnpm-workspace.yaml
NicoSerranoP: Same answer as above. The `unrs-resolver` is not rebuild in local and the `eslint` package is rebuild in local because it needs different libraries/utilities depending on the OS

PR #23, opened 2026-02-23
feat: benchmark against time window

JohnGuilding:
### What does this PR do?
  - Scan full a 7-day block window instead of sampling a few hundred receipts
      - Use `BLOCK_WINDOW_ETHEREUM` & `BLOCK_WINDOW_SCROLL` for defining 7 day window
  - Remove `MAX_OF_LOGS` per protocol. This is because we are using a time window based approach. We can keep an eye on Alchemy api key usage and optimise if requests become too cost prohibitive. The account already has a max usage policy to protect against very high usage.  
  - Remove `NUMBER_OF_TRANSACTIONS` for same reasons as above
  - Add `MIN_SAMPLES` = 50. This is a minimum floor we want for averages. Maybe 50 could be too high, as some protocols might not get a lot of use at all. In which case, we could remove, adjust this later
  - Add `MAX_SAMPLES` = 100,000. This is a safety guard to prevent us from sending extremely high amounts of rpc requests
  - Increase `BLOCK_RANGE` to 2000. It appears 2000 is a [safe option](https://www.alchemy.com/docs/chains/ethereum/ethereum-api-endpoints/eth-get-logs) for `getLogs`
  - Merge `getEventLogs` + `getTransactionsWithEvents` into single
  `getValidTransactions` function. 
  - `getValidTransactions` calls `getAllLogs` & `getValidReceipts`
      - `getAllLogs` — scans blocks in chunks from newest to oldest, deduplicates by
  tx hash in a single pass using an inline Set
    - `getValidReceipts` — fetches receipts and filters to those whose log count
  and event topics match the expected pattern
  - Deduplicate logs in a single pass using inline Set instead of separate
  `getUniqueLogs` calls
  - Make choice of `chain` and explicit choice in order to use correct client and block window logic. This becomes more important as we add more chains (e.g. protocol only deployed on Base)
- Add minor natspec comments
- Move stable env vars definition to constants
- Add env var validation
  
### Some changes to simplify logic
  - Remove `reduce` and use for..of
  - Move `getBlockInRange` and `getFromAndToBlocks` inline
  - Move types and utils defined once inline if it didn't add to LoC too much

### Why do we want to scan by a consistent block window?

The main reason we want to do this is because if we get 200 receipts per protocol, that will mean very different things depending on protocol usage. If protocol A has 200 valid txs in one week, whereas protocol B has 20,000 valid txs in one week, then we are not getting an equivalent average of both. The average of the protocol with 200 in week will encompass gas prices from 7 whole days, whereas protocol B might only have an average based on a day of usage.

This way, protocol benchmarks reflect the same range in gas prices, and subsequent changes that has on utilisation (gas prices may impact how much people use the chain).

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -28,25 +28,18 @@
 - Exception: loops acceptable for complex control flow (see `src/utils/rpc.ts` pagination)
 - Examples: `src/utils/utils.ts`, `src/utils/rpc.ts`, `src/index.ts`
 
-## 4. Type Organization
-
-- **ALL** types/interfaces MUST be in dedicated `types.ts` files
-- **NEVER** define types inline in implementation files
-- Locations: `src/utils/types.ts` (shared), `src/__tests__/types.ts` (test-specific)
-- Use inline type imports: `import type { Type } from "./types.js";`
-
-## 5. Utility Function Organization
+## 4. Utility Function Organization
 
 - **ALL** reusable functions MUST be in dedicated utility files
 - **NEVER** define utilities inline with implementation code
 - Locations: `src/utils/utils.ts`, `src/utils/rpc.ts`, `src/utils/db.ts`, `src/__tests__/utils.ts`
 - Prefer pure functions that minimize side effects
 
-## 6. Clean Code Principles
+## 5. Clean Code Principles
 
 - Follow Clean Code guidelines, but rules 1-4 above take precedence if there's a conflict
 
-## 7. Testing Requirements
+## 6. Testing Requirements
 
 - **ALWAYS** run `pnpm run test` and `pnpm run benchmark` after changes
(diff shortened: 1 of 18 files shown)


NicoSerranoP (approved):
I love it! 


PS: never mind I just saw the next PR 🚀 

~~One non-blocking discussion that we can implement later is: `for` vs `reduce`. In the `client.getTransactionReceipt(...)` I think it would be better to sacrifice readability for performance by using `reduce` instead of `for`.~~

~~Right now the protocol specific benchmarking promise thread has to wait `MAX_SAMPLES` loop iterations. In reduce these iterations will be ran concurrently and the fetching will stop after reaching `MAX_SAMPLES`. The iteration will continue but they will break before the `await client.getTransactionReceipt(...)`~~

PR #27, opened 2026-02-24
Add Hinkal protocol gas benchmarking

### LINKED ISSUE #21: Benchmark Hinkal

NicoSerranoP:
## Summary
This PR adds gas benchmarking support for the Hinkal privacy protocol, enabling measurement of gas costs for shield, unshield, and private transfer operations on ERC20 tokens.

## Key Changes
- **New Hinkal benchmarking module** (`gas-benchmarks/src/hinkal/`):
  - `index.ts`: Implements the `Hinkal` class with three benchmark methods:
    - `benchmarkShieldERC20()`: Measures gas for depositing ERC20 tokens into the Hinkal pool
    - `benchmarkUnshieldERC20()`: Measures gas for withdrawing ERC20 tokens from the pool
    - `benchmarkTransferERC20()`: Measures gas for private transfers within the pool
  - `constants.ts`: Defines the Hinkal Pool contract address and event signatures for tracking shield, unshield, and transfer operations

- **Integration into main benchmarking pipeline** (`gas-benchmarks/src/index.ts`):
  - Added Hinkal instance initialization
  - Integrated Hinkal benchmarking into the parallel benchmark execution
  - Added Hinkal metrics to the database update with properly formatted keys

## Implementation Details
- The Hinkal benchmarking follows the same pattern as existing protocols (Railgun, Tornado Cash, etc.)
- Uses event-based transaction filtering to identify relevant operations:
  - Shield operations tracked via `Transfer` and `NewCommitment` events
  - Unshield operations tracked via `Transfer` and `Nullified` events
  - Private transfers tracked via `Nullified` and `NewCommitment` events
- Validates minimum sample size (MIN_SAMPLES) before calculating average metrics
- Targets the Hinkal Pool contract at `0x25e5e82f5702A27C3466fE68f14abDbbAdFca826` (consistent across all supported chains)

https://claude.ai/code/session_01PFNChP2EtEQ4AUMXD7E1zZ

Closes #21

### DIFF
--- a/gas-benchmarks/src/hinkal/constants.ts
+++ b/gas-benchmarks/src/hinkal/constants.ts
@@ -1,0 +1,93 @@
+import { parseAbiItem, type Address } from "viem";
+
+/**
+ * Hinkal Pool contract deployed at the same address across all supported chains
+ * (Ethereum, Arbitrum, Optimism, Polygon, Base):
+ * https://github.com/Hinkal-Protocol/Hinkal-Smart-Contracts/blob/main/contracts/Hinkal.sol
+ */
+export const HINKAL_POOL: Address = "0x25e5e82f5702A27C3466fE68f14abDbbAdFca826";
+
+/**
+ * Hinkal.transact function - shield native ETH into the Hinkal pool. It does not emit outputs (nullified):
+ * https://github.com/Hinkal-Protocol/Hinkal-Smart-Contracts/blob/8bd371883a0d8f6f56775228ee48a1bbf9fecbac/contracts/Hinkal.sol#L40
+ * The repo has been close sourced since March 31th 2026
+ *
+ * Emits:
+ * NewCommitment() - New private commitment inserted into the Merkle tree (emitted by HinkalBase.insertCommitments). A deposit has 1 commitment and 0 nullified events
+ *
+ * Example:
+ * https://etherscan.io/tx/0x90fe99c5df0f36ace8a2ab3ff2b03e147413374eb28737619d41f2b920c15112
+ *
+ * Note: deposits can be done with .transact() and also with .prooflessDeposit() function execution
+ */
+export const SHIELD_ETH_EVENTS = [
+  parseAbiItem("event NewCommitment(uint256 commitment, int256 index, bytes encryptedOutput)"),
+] as const;
+
+/**
+ * Hinkal.transact function (unshield) - withdraw native ETH from the Hinkal pool. It does not outputs new notes (commitments)
+ * https://github.com/Hinkal-Protocol/Hinkal-Smart-Contracts/blob/8bd371883a0d8f6f56775228ee48a1bbf9fecbac/contracts/Hinkal.sol#L40
+ * The repo has been close sourced since March 31th 2026
+ *
+ * Emits:
+ * Nullified() - Input commitment nullifier marked as spent (emitted by HinkalBase.insertNullifiers)
+ *
+ * Example:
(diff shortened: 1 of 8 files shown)


gas-benchmarks/src/index.ts
```diff
@@ -53,6 +57,16 @@ await db.update((data) => {
...
   };
+
+  // eslint-disable-next-line no-param-reassign
+  data[`${hinkal.name}_${hinkal.version}`] = {
+    shield_eth: hinkalMetrics.shieldEth,
+    unshield_eth: hinkalMetrics.unshieldEth,
+    internal_transfer: hinkalMetrics.internalTransfer,
+    shield_erc20: hinkalMetrics.shieldErc20,
+    unshield_erc20: hinkalMetrics.unshieldErc20,
+    transfer_erc20: hinkalMetrics.transferErc20,
```
JohnGuilding: small optimisation, why don't we re-use result of `internalTransfer` for `transfersErc20` if it is the same thing?

gas-benchmarks/src/hinkal/constants.ts
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * Emits:
+ * NewCommitment() - New private commitment inserted into the Merkle tree (emitted by HinkalBase.insertCommitments). A deposit has 1 commitment and 0 nullified events
+ *
+ * Example:
+ * https://etherscan.io/tx/0x90fe99c5df0f36ace8a2ab3ff2b03e147413374eb28737619d41f2b920c15112
+ *
+ * Note: deposits can be done with .transact() and also with .prooflessDeposit() function execution
+ */
+export const SHIELD_ETH_EVENTS = [
```
JohnGuilding: General question on benchmarks, I'm wondering if being strict on the exact events emitted is having any negative effect on sample accuracy.

For example, the sample set could be artificially small and potentially biased toward a specific transaction pattern when we only use a specific combination of events.

Do you think there is a way to test this? E.g. could be spot check one or two protocols manually to verify our methodology. E.g. we check that privacy pools actually costs xxxx gas. 

I'm uncertain if it's an issue, or if there is a good way to test it. Perhaps spot checking our benchmarks is a reasonable thing to do anyway before we publish this externally

NicoSerranoP: I have been thinking about this as well. In my opinion in order to make the comparison as similar as possible, we should try to focus only on cheaper/simple txs trying to represent an atomic private transfer.

Could we discuss the spot checking methodology in this [issue](https://github.com/privacy-ethereum/private-transfers-benchmarks/issues/144) and in following meetings?

gas-benchmarks/src/hinkal/constants.ts
```diff
@@ -0,0 +1,93 @@
...
+] as const;
+
+/**
+ * Hinkal.transact function (unshield) - withdraw ERC20 tokens from the Hinkal pool:
+ * https://github.com/Hinkal-Protocol/Hinkal-Smart-Contracts/blob/8bd371883a0d8f6f56775228ee48a1bbf9fecbac/contracts/Hinkal.sol#L40
+ * The repo has been close sourced since March 31th 2026
+ *
+ * Emits:
+ * Transfer() - ERC20 token transfer from Hinkal pool to recipient (emitted by the ERC20 token contract)
+ * Transfer() - ERC20 token transfer from Hinkal pool to relayer (emitted by the ERC20 token contract, only for unshielding with relayer fee)
```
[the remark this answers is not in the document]
JohnGuilding: ohhhh interesting

PR #29, opened 2026-03-04
feat(monorepo): add project evaluations

NicoSerranoP:
1. Move all TS configuration to mono repo root in order to have a single source of truth for ESLint and Prettier
2. Add a project-evaluations directory to modify data.json


NicoSerranoP:
@JohnGuilding I am been using the edit functionality to put the values. It is easier than modifying the JSON itself

PR #142, opened 2026-03-30
feat: add zcash analysis

JohnGuilding:
- Add zcash analysis using research script
- Modified some answers manually to increase accuracy
- 25 properties have citations with `char_location` references to source documents
- 5 properties flagged `needsResearchReview` where no citable source was found (values based on model   knowledge)                                                                                            
- Does not research Anonymity set size and gas costs, as that requires benchmark data
- Show zcash and citations in frontend

This use the research script that calls the Anthropic API with the citation feature. A few properties do not have citations because the right url could not be found, so it relied on training data and requires additional manual review. Indicated by the `"needsResearchReview": true` flag

### DIFF
--- a/project-evaluations/scripts/research-prompts.ts
+++ b/project-evaluations/scripts/research-prompts.ts
@@ -46,7 +46,7 @@ const BASE_RULES = `
   SOURCE SELECTION RULES:
   21. The page must contain specific technical content about this property — not just a table of contents, navigation page, or landing page.
   22. Do NOT return PDF URLs — they cannot be processed. Find an HTML alternative.
-  23. Avoid blog posts as primary sources — they may be outdated. Prefer official docs, specs, or project websites.
+  23. Never use blog posts from exchanges, trading platforms, or non-official sources. Only use blog posts from the protocol's own team or foundation (e.g. electriccoin.co/blog for Zcash, blog.ethereum.org for Ethereum). Prefer official docs, specs, or project websites over any blog.
   24. If the best source is a PDF, search for an HTML version of the same content.`;
 
 // prettier-ignore


project-evaluations/src/data/evaluations/zcash.json
```diff
@@ -0,0 +1,596 @@
...
+        {
+          "cited_text": "Therefore, there are four basic types of transactions:\n\nShielded/private (Value is not revealed on the blockchain) ¶ \n\nDeshielding (Value is revealed on the receiver end) ¶ \n\nShielding (Value is revealed on the sender end) ¶ \n\nTransparent/public (Value is revealed by both sender and receiver) ¶ \n\nShielded Addresses ¶ \n\nShielded addresses are the address type that use zero-knowledge proofs to allow transaction data to be encrypted but remain verifiable by network nodes. ",
+          "source": "https://zcash.readthedocs.io/en/latest/rtd_pages/addresses.html",
+          "start_char_index": 1325,
+          "end_char_index": 1799
+        },
+        {
+          "cited_text": "Shielded addresses and the values sent to or from them are not publicly visible. ¶ \n\nA transaction between two shielded addresses (a shielded transaction) keeps the addresses, transaction amount and the memo field shielded from the public (with the exception of migrating funds between Sprout and Sapling shielded addresses).\n\n",
+          "source": "https://zcash.readthedocs.io/en/latest/rtd_pages/addresses.html",
+          "start_char_index": 2105,
```
NicoSerranoP: Are `"start_char_index"` and `"end_char_index"` needed in the report? From what I saw from the source (https://zcash.readthedocs.io/en/latest/rtd_pages/addresses.html), there is no easy way to count the characters and find the specific place where the citation was found. 

I would suggest getting the most approximate link (e.g. https://zcash.readthedocs.io/en/latest/rtd_pages/addresses.html#id6) but this should be done by a human. It would be super subjective for a machine

project-evaluations/src/data/evaluations/zcash.json
```diff
@@ -0,0 +1,596 @@
...
+    {
+      "name": "On-chain gas cost: transfer",
+      "value": ""
+    },
+    {
+      "name": "On-chain gas cost: withdraw",
+      "value": ""
+    },
+    {
+      "name": "Time-to-finality",
```
NicoSerranoP: I changed time-to-finality metric to seconds in order to have all the times in the same metric (time-to-finality, deposit and withdraw). I pushed a commit with grammar fixes and these conversions :)

project-evaluations/src/data/evaluations/zcash.json
```diff
@@ -0,0 +1,596 @@
...
+          "start_char_index": 2336,
+          "end_char_index": 2748
+        }
+      ]
+    },
+    {
+      "name": "Point of enforcement",
+      "value": "[\"None\"]",
+      "notes": "Zcash has no protocol-level compliance enforcement at any stage. The protocol supports both transparent and shielded addresses with optional privacy, allowing users to freely move funds between pools through shielding and deshielding transactions. All transactions are validated by consensus rules using zero-knowledge proofs for shielded transfers, but there is no active restriction or filtering mechanism enforced by the protocol.",
+      "citations": [
```
NicoSerranoP: I feel that citations in Point of enforcement are not related to the property. We could use the citation in Type of compliance

JohnGuilding: Agreed, swapping it to another citation that mentions viewing keys. Using some of the same citations multiple times, but think that is ok if it proves the point

project-evaluations/src/data/evaluations/zcash.json
```diff
@@ -0,0 +1,596 @@
...
+      "name": "Confidentiality",
+      "value": "Yes",
+      "notes": "Zcash provides full confidentiality for shielded addresses, where both balances and transaction amounts remain private through zero-knowledge proofs. A transaction between two shielded addresses keeps the addresses, transaction amount, and memo field shielded from the public. Transparent addresses work similarly to Bitcoin addresses and do not offer privacy, with addresses and values publicly visible.",
+      "citations": [
+        {
+          "cited_text": "Therefore, there are four basic types of transactions:\n\nShielded/private (Value is not revealed on the blockchain) ¶ \n\nDeshielding (Value is revealed on the receiver end) ¶ \n\nShielding (Value is revealed on the sender end) ¶ \n\nTransparent/public (Value is revealed by both sender and receiver) ¶ \n\nShielded Addresses ¶ \n\nShielded addresses are the address type that use zero-knowledge proofs to allow transaction data to be encrypted but remain verifiable by network nodes. ",
+          "source": "https://zcash.readthedocs.io/en/latest/rtd_pages/addresses.html",
+          "start_char_index": 1325,
+          "end_char_index": 1799
+        },
```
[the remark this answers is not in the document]
JohnGuilding: So I didn't change `url` since I didn't want to mess with your analysis and it looked like that was specifically for documentation, whereas citation urls are broader.

A future improvement I was thinking of: we might want to allow the research script to add multiple urls, currently it only supports 1. So citation[0] references one source, but citation[1] might reference another. This would be a simple experiment by just changing the prompts to return two source urls instead of one.

I would like to test out the citation feature a bit more before committing to it, so initial suggestion would be to leave both url and citations as optional, and then migrate properly once we have certainty of the better approach, and we have tried out using multiple sources?

project-evaluations/src/data/evaluations/zcash.json
JohnGuilding: on second thought, one feature is that we could update the UI to link to the cited text and highlight it for the reviewer, that would make the UX of reviewing in the UI quite easy, you just click once, and then you see the text highlighted, this is slightly easier than doing copying, then running cmd+f. It's a one click review per citation. 

So since the API gives us this for free, and this could be a feature to experiment with, would say we could keep the optionality for now, and delete later if we don't need? Wdyt?

JohnGuilding: alright sounds gooood

NicoSerranoP (approved):

PR #154, opened 2026-04-09
feat: add anonymity set size benchmark

### LINKED ISSUE #156: Add anonymity set size benchmark

0xmad:
Add anonymity set size benchmark for all the gas benchmarked projects

### DIFF
--- a/gas-benchmarks/src/privacy-pools/index.ts
+++ b/gas-benchmarks/src/privacy-pools/index.ts
@@ -1,6 +1,7 @@
 import { mainnet } from "viem/chains";
 
-import type { FeeMetrics } from "../utils/types.js";
+import type { IAggregatedMetrics } from "./types.js";
+import type { TransactionReceipt } from "viem";
 
 import { MIN_SAMPLES } from "../utils/constants.js";
 import { getValidTransactions } from "../utils/rpc.js";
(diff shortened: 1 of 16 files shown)


JohnGuilding:
@0xmad why are we doing `shieldEthReceipts.length - unshieldEthReceipts.length` for anonymity set size?

0xmad:
Because we want to distinguish how many private deposits left in the pool. Withdrawals can be calculated as well because they are not part of the pool anymore.

gas-benchmarks/src/privacy-pools/index.ts
```diff
@@ -18,13 +19,20 @@ export class PrivacyPools {
...
+  async benchmark(): Promise<IAggregatedMetrics> {
+    const [shieldEthReceipts, unshieldEthReceipts] = await Promise.all([
+      this.benchmarkShieldETH(),
+      this.benchmarkUnshieldETH(),
+    ]);
+
+    return {
+      shieldEth: getAverageMetrics(shieldEthReceipts),
+      unshieldEth: getAverageMetrics(unshieldEthReceipts),
+      anonymitySetSize: shieldEthReceipts.length - unshieldEthReceipts.length,
```
JohnGuilding, line 31: John:
> @0xmad why are we doing shieldEthReceipts.length - unshieldEthReceipts.length for anonymity set size?

Anton:
> Because we want to distinguish how many private deposits left in the pool. Withdrawals can be calculated as well because they are not part of the pool anymore.

So we actually want to get to [total size of the anonymity set](https://www.notion.so/pse-team/State-of-Private-Transfers-2efd57e8dd7e80cdb60ed12f7e4913a7?source=copy_link#2efd57e8dd7e806b8dddfe0ebd9c4390), not the size within a specific window. If we sample 5 weeks of blocks, we will get the net change within that window, but not the cumulative anonymity set size

JohnGuilding (changes requested):

gas-benchmarks/src/privacy-pools/index.ts
0xmad: With current design, it’s not possible to fetch all the events and calculate total size of deposits.

JohnGuilding: Yes definitely. So I think we could  also try measure anonymity set over a time window, but our operation (shield, transfer, unshield) capturing logic only measures simple transfers for many protocols. So the "unspent" anonymity set window would not account for all relevant state updates. E.g we are measuring transfers with 1 input and 1 output, but not transfers with 1 input and 2 outputs.

This is either too complex to not be worthwhile, or requires something quite involved. I'm going to leave some notes in [this issue](https://github.com/privacy-ethereum/private-transfers-issues/issues/96)

PR #157, opened 2026-04-14
feat: railgun subgraph for shield and unshield events

0xmad:
- [x] Added subgraph boilerplate
- [x] Implemented event handlers for Shield and Unshield events
- [x] Added tests

### DIFF
--- a/subgraph/schema.graphql
+++ b/subgraph/schema.graphql
@@ -1,0 +1,137 @@
+type RailgunProtocolStats @entity(immutable: false) {
+  id: ID!
+  totalTxCount: BigInt!
+  totalGasUsed: BigInt!
+
+  shield: RailgunOperationStats!
+  unshield: RailgunOperationStats!
+}
+
+type RailgunShieldTokenStats @entity(immutable: false) {
+  id: ID!
+  tokenAddress: Bytes!
+  totalCount: BigInt!
+  totalGasUsed: BigInt!
+  totalValue: BigInt!
+
+  operationStats: RailgunOperationStats!
+}
+
+type RailgunUnshieldTokenStats @entity(immutable: false) {
+  id: ID!
+  tokenAddress: Bytes!
+  totalCount: BigInt!
+  totalGasUsed: BigInt!
+  totalValue: BigInt!
+
+  operationStats: RailgunOperationStats!
+}
+
+type RailgunOperationStats @entity(immutable: false) {
+  id: ID!
+  totalCount: BigInt!
+  totalGasUsed: BigInt!
+
+  shieldTokenStats: [RailgunShieldTokenStats!]! @derivedFrom(field: "operationStats")
+  unshieldTokenStats: [RailgunUnshieldTokenStats!]! @derivedFrom(field: "operationStats")
+}
+
+type RailgunShield @entity(immutable: true) {
+  id: ID!
+  treeNumber: BigInt!
+  startPosition: BigInt!
+  fees: [BigInt!]!
+
+  commitments: [RailgunCommitment!]! @derivedFrom(field: "shield")
+  ciphertexts: [RailgunShieldCiphertext!]! @derivedFrom(field: "shield")
+
+  blockNumber: BigInt!
+  timestamp: BigInt!
+  txHash: Bytes!
+  gasUsed: BigInt!
+  gasPrice: BigInt!
+}
+
+type RailgunCommitment @entity(immutable: true) {
+  id: ID!
+  shield: RailgunShield!
+
+  npk: Bytes!
+
+  tokenType: Int!
+  tokenAddress: Bytes!
+  tokenSubID: BigInt!
+
+  value: BigInt!
+}
+
+type RailgunShieldCiphertext @entity(immutable: true) {
+  id: ID!
+  shield: RailgunShield!
+
+  encryptedBundle0: Bytes!
+  encryptedBundle1: Bytes!
+  encryptedBundle2: Bytes!
+
+  shieldKey: Bytes!
+}
+
+type RailgunNullified @entity(immutable: true) {
+  id: ID!
+  treeNumber: Int!
+
+  nullifiers: [Bytes!]!
+
+  blockNumber: BigInt!
+  timestamp: BigInt!
+  txHash: Bytes!
+  gasUsed: BigInt!
+}
+
+type RailgunUnshield @entity(immutable: true) {
+  id: ID!
+  to: Bytes!
+
+  tokenType: Int!
+  tokenAddress: Bytes!
+  tokenSubID: BigInt!
+
+  amount: BigInt!
+  fee: BigInt!
+
+  blockNumber: BigInt!
+  timestamp: BigInt!
+  txHash: Bytes!
+  gasUsed: BigInt!
+  gasPrice: BigInt!
+}
+
+type RailgunTransact @entity(immutable: true) {
+  id: ID!
+  treeNumber: BigInt!
+  startPosition: BigInt!
+  hash: [Bytes!]!
+
+  ciphertexts: [RailgunTransactCiphertext!]! @derivedFrom(field: "transact")
+
+  blockNumber: BigInt!
+  timestamp: BigInt!
+  txHash: Bytes!
+  gasUsed: BigInt!
+}
+
+type RailgunTransactCiphertext @entity(immutable: true) {
+  id: ID!
+  transact: RailgunTransact!
+
+  ciphertext0: Bytes!
+  ciphertext1: Bytes!
+  ciphertext2: Bytes!
+  ciphertext3: Bytes!
+
+  blindedSenderViewingKey: Bytes!
+  blindedReceiverViewingKey: Bytes!
+
+  annotationData: Bytes!
+  memo: Bytes!
+}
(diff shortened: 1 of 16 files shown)


NicoSerranoP (commented):
I like this approach, I was testing it and playing with it in my local to understand more how it works. I will leave a couple of comments 😊

subgraph/src/railgun-smart-wallet.ts
```diff
@@ -0,0 +1,214 @@
...
+  stats.save();
+
+  shield.treeNumber = event.params.treeNumber;
+  shield.startPosition = event.params.startPosition;
+  shield.fees = event.params.fees;
+
+  shield.blockNumber = event.block.number;
+  shield.timestamp = event.block.timestamp;
+  shield.txHash = event.transaction.hash;
+  shield.gasUsed = event.transaction.gasLimit;
```
NicoSerranoP: Can we save the `transaction.effectiveGasPrice` so we have a final value that the user is paying for each tx?

subgraph/src/railgun-smart-wallet.ts
```diff
@@ -0,0 +1,214 @@
...
+
+  for (let index = 0; index < event.params.commitments.length; index += 1) {
```
NicoSerranoP, line 119: I think there are 3 types of txs with commitments:

- deposit transactions (generate commitments)
- internal private transfer txs (generate commitments for the receiver to use later)
- DeFi interactions txs (generate commitments to claim funds later)


How are we going to distinguish between the 3? I can see a schema defined as:

```
type RailgunShield @entity(immutable: true) {
  id: ID!
  treeNumber: BigInt!
  startPosition: BigInt!
  fees: [BigInt!]!

  commitments: [RailgunCommitment!]! @derivedFrom(field: "shield")
  ciphertexts: [RailgunShieldCiphertext!]! @derivedFrom(field: "shield")

  blockNumber: BigInt!
  timestamp: BigInt!
  txHash: Bytes!
  gasUsed: BigInt!
}
```

but I am not 100% sure how we are separating Shield txs from other type txs

subgraph/src/railgun-smart-wallet.ts
```diff
@@ -0,0 +1,214 @@
...
+      tokenStats.totalGasUsed = tokenStats.totalGasUsed.plus(gasPerCommitment);
+      tokenStats.totalValue = tokenStats.totalValue.plus(eventCommitment.value);
+
+      tokenStats.save();
+    }
+
+    commitment.save();
+  }
+
+  const ciphertexts = event.params.shieldCiphertext;
```
NicoSerranoP, line 152: Not sure if we need to collect ciphertexts 😊 . What do you think?

subgraph/tests/railgun-smart-wallet-shield.test.ts
NicoSerranoP, line 1: These tests emulate events based on the handlers parameters. Would there be a way to run a test on production data? Like deploying the subgraph and querying to see if the metrics have values?

subgraph/networks.json
NicoSerranoP, line 1: Why do we need the `networks.json` file and the `subgraph.yaml` file? they both specify the Railgun contract address and deployment block but I was wondering if there is a way to have a single source of truth? It will make it easy to maintain all contract addresses and deployment blocks in one place rather than in 2

subgraph/subgraph.yaml
```diff
@@ -0,0 +1,35 @@
...
+      language: wasm/assemblyscript
+      entities:
+        - Shield
+        - Nullified
+        - Unshield
+        - Transact
+      abis:
+        - name: RailgunSmartWallet
+          file: ./abis/RailgunSmartWallet.json
+      eventHandlers:
```
NicoSerranoP, line 26: I am a bit lost: what does `eventHandlers` do? Do we defined the events we are looking here? What happens if the deposit action/tx emits 2 events: `Transfer` and `Shield`? Can we put these 2 here and the subgraph will index only txs with these 2 events in sequential order?

subgraph/schema.graphql
NicoSerranoP, line 1: Is there a way to divide this `schema.graphql` file into different files: `railgun/schema.graphql`, `privacy-pools/schema.graphql`? We could have the data in different files and have a script to merge it into a single `schema.graphql` before running `pnpm run codegen`

NicoSerranoP (commented):
I like the approach! Left a couple of comments to discuss them here. I think that if this is the way we are going to go I need to understand more how does it work / how is it structured

subgraph/src/railgun-smart-wallet.ts
NicoSerranoP: Why are we using `transaction.gasLimit` as `gasUsed`? I thought `gasLimit` was how much the user is willing to spend while `gasUsed` is how much it was actually spent

subgraph/src/railgun-smart-wallet.ts
0xmad:
```
type RailgunCommitment @entity(immutable: true) {
  id: ID!
  shield: RailgunShield!

  npk: Bytes!

  tokenType: Int!
  tokenAddress: Bytes!
  tokenSubID: BigInt!

  value: BigInt!
}
```
This is commitment struct, there is no type for transactions and shield is not the same as transact or unshield.

subgraph/src/railgun-smart-wallet.ts
0xmad: This data is in shield event

subgraph/tests/railgun-smart-wallet-shield.test.ts
0xmad: It's not possible to do with matchstick, it's a different test set (e2e)

subgraph/subgraph.yaml
0xmad: We handle the events, not transactions. Both events will be mapped with the same transaction hash.

subgraph/schema.graphql
0xmad: It's possible but not for now

subgraph/src/railgun-smart-wallet.ts
0xmad: Receipts are not available in events and gasLimit is approx. the same as gasUsed.

0xmad: Ok, so there is a setting where reciept can be fetched as well

subgraph/src/railgun-smart-wallet.ts
NicoSerranoP: Got it! So we need to collect all data types emitted by events. I understand it more now 😊

subgraph/src/railgun-smart-wallet.ts
NicoSerranoP: Got it! I think this question is related to the 2 consecutive events question below. I will close it

subgraph/tests/railgun-smart-wallet-shield.test.ts
NicoSerranoP: Got it! These tests are important to verify the handlers are working correctly but the emulated events might make us overlook something (we are creating the test data that will be used to verify our own written handlers). I think we should use e2e tests with production on-chain data in the future so we can guarantee that the handlers are working.

I will leave it as an [open issue](https://github.com/privacy-ethereum/private-transfers-benchmarks/issues/158) for future work

subgraph/subgraph.yaml
NicoSerranoP: Got it! How should we proceed in cases such as Hinkal (I have the codebase if you need it. They made it private last month):

1. **Shield** function call emits `NewCommitment` event (only one because you are just creating a note)

2. **Unshield** function call emits `Nullified` event (can be multiples times if user is spending multiple notes. We decided to just benchmark txs with single Nullified events to have the atomic price of a single unshield action)

3. **Internal transfers**  function call emits multiple `NewCommitment` and `Nullified` events depending on the performed action (internal transfer, DeFi interaction, etc).


All these transactions are performed by calling the `transact` function in the Hinkal smart contract.

Check our current benchmarking constants: https://github.com/privacy-ethereum/private-transfers-benchmarks/blob/main/gas-benchmarks/src/hinkal/constants.ts

Check their docs: https://hinkal-team.gitbook.io/hinkal/technical-description/smart-contracts

subgraph/src/railgun-smart-wallet.ts
NicoSerranoP: I think that depends a lot on the user/wallets. I have seen some cases where wallets put a high gas limit and you end up using a lot less. Claude gave me these 3 bulletpoints:

- ERC-20 transfers typically use ~21k–65k gas but senders often set limits of 100k–200k+ as a safety buffer
- **Not applicable in our case:** ~Failed transactions use very little gas (just enough to hit the revert) but consume their full gasLimit from the sender's perspective — using gasLimit would massively overestimate cost~
- Actual gas efficiency of your contract implementation is only measurable via gasUsed

Is there a way to fetch that data from subgraph? We can leave it as future work for another PR :)

subgraph/subgraph.yaml
0xmad: Transaction can emit multiple events and on subgraph side it's possible to parse these events and group them by transaction hash.

NicoSerranoP: Cool! That will work :)

NicoSerranoP (approved):
Sorry I messed up with the commit. At the moment we have the latest pushes (gasUsed+ gasPrice) 🚀

PR #159, opened 2026-04-16
feat: add railgun transact event handler

0xmad:


subgraph/src/railgun-smart-wallet.ts
```diff
@@ -32,9 +39,14 @@ function createOrLoadProtocolStats(): RailgunProtocolStats {
     unshieldStats.totalCount = BigInt.fromI32(0);
     unshieldStats.totalGasUsed = BigInt.fromI32(0);
 
+    stats.transact = transactStats.id;
```
NicoSerranoP, line 42: Nitpick: should we replace `id` by `name` in the OperationStats schemas to make it more readable? I went back to the code to check what id was 😂

subgraph/src/railgun-smart-wallet.ts
```diff
@@ -233,3 +241,68 @@ export function handleUnshield(event: UnshieldEvent): void {
...
+    stats.totalGasUsed = stats.totalGasUsed.plus(event.receipt!.gasUsed);
+  }
+
+  const transactStats = RailgunOperationStats.load(stats.transact);
+
+  if (transactStats !== null) {
+    transactStats.totalCount = transactStats.totalCount.plus(BigInt.fromI32(1));
+
+    if (event.receipt !== null) {
+      transactStats.totalGasUsed = transactStats.totalGasUsed.plus(event.receipt!.gasUsed);
```
NicoSerranoP, line 263: `RailgunSmartWallet.transact()` can execute batch of Railgun txs. How are we filtering out so that we measure the correct value of sending a private transfer? Some txs will have 2 txs, others will have 1 tx. If we just sum up the gasUsed without filtering the Ethereum txs by 1 Railgun tx we will be saving the wrong metric

NicoSerranoP (commented):
Looks good! Could we implement the transaction accumulator to check how many other events have been emitted in the Transact tx?

subgraph/src/railgun-smart-wallet.ts
0xmad: Entities are linked together by id field, it's not possible to change it.

subgraph/src/railgun-smart-wallet.ts
0xmad: Added different total count for shield and transact

NicoSerranoP: Got it! We will consider every commit a new transaction.

NicoSerranoP (approved):

PR #160, opened 2026-04-17
feat(benchmarks): add redact

NicoSerranoP:
This PR adds the Redact (confidential token project built with Fhenix) to the current benchmarking methodology. I have found the following issues:

- I am having a lot of RPC rate limit errors. I am batching the concurrent requests to reduce the number of request per time period
- I also activated the strict mode in `client.getLogs()` but it does not seem to be doing anything useful in our case:

we assumed that `getLogs({ events: [...] })` will filter transactions with a specific log order but it does not work like that. It fetches all logs emitted from the specified address that adhere to any of the events structure. This makes it difficult to differentiate txs. For example in redact

1. **Shield:** send public ETH to contract and get encrypted ETH (eETH). This tx emits 10 events: 7 TaskCreated (for the Fhenix coprocessor), 1 Transfer, 1 ConfidentialTransfer and 1 ShieldedNative.
2. **Unshield:** send encrypted ETH (eETH) to public ETH. This tx emits 10 events: 7 TaskCreated (for the Fhenix coprocessor), 1 Transfer, 1 ConfidentialTransfer and 1 Unshielded. Worth pointing out that the unshield process involves Unshield and also Claim.
3. **Transfer:** send encrypted ETH (eETH) to another address. This tx emits 8 events: 6 TaskCreated (for the Fhenix coprocessor), 1 Transfer and 1 ConfidentialTransfer

As you can see all 3 types of tx emit ConfidentialTransfer and the only way to differentiate a **Transfer** from the other 2 is by counting the emitted events and checking it does not contain ShieldedNative or Unshielded.

We can merge this PR as reference and keep the above idea in mind when migrating to the subgraph methodology

### DIFF
--- a/gas-benchmarks/.env.example
+++ b/gas-benchmarks/.env.example
@@ -2,6 +2,8 @@ ETH_RPC_URL=https://mainnet.gateway.tenderly.co
 
 SCROLL_RPC_URL=https://scroll.gateway.tenderly.co
 
+SEPOLIA_RPC_URL=https://sepolia.gateway.tenderly.co
+
 MONERO_FAIL_NODES_API_URL=https://monero.fail/nodes.json
 
 MIN_SAMPLES=10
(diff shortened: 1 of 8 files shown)


0xmad (commented):
@NicoSerranoP why do we continue using old way for benchmark instead of adding event parsing in subgraph?

PR #163, opened 2026-04-21
feat: update research script

JohnGuilding:
**What this PR does:**
Restructure the research script into a multi-phase flow using Claude code skills:
1. Add the protocol to `scripts/research-config.ts`.
2. Run skill: `/research-sources <id>` in Claude Code to write `scripts/research-cache/<id>.json`.
3. Run script: `pnpm run research <id>` to write `src/data/evaluations/<id>.json`.
4. Run skill: `/review-evaluation <id>` to check against the rules in `scripts/research-prompts.ts`.
5. Prompt rules are automatically updated with review feedback

Above steps are composed into single skill:
`/evaluate-protocol <id>`

This reduces token usage significantly since the API is used only for citation feature.

- Introduce the new flow: `/research-sources` discovers URLs and writes `scripts/research-cache/{id}.json`, `scripts/research-protocol.ts` reads the cache, fetches each URL once, and calls the Anthropic API per property to get citations
- Switch the API call to the tool-use API. Claude now writes the note as a plain text block and records the structured value via a separate tool call (`record_evaluation`). Separating the two outputs gives the Citations API a clean text channel to attach source quotes to. The rule changes such as guidance on paraphrasing meant that citation feature stopped working, this changes makes citations more robust
- Enable prompt caching on the document block (document blocks needed by citation feature). Fetched pages are multiple KB each and previously re-sent per property. Caching the document means only the first per-URL call pays the full input cost, and the shared URLs across properties (5-10 unique URLs per 30-property run) hit the cache for the rest.
- Add many new rules in `research-prompts.ts`
- Rewrite rules in `research-prompts.ts`: reframe Cryptographic verifiability, fix Number of secrets minimum, add Selective-disclosure carveout to CROSS_CHECK_RULES, tighten WRITING/VALUE_FORMAT/PROPERTY rules
- Expand `research-config.ts`: add context field to ProtocolConfig and add Nullmask/Houdiniswap/Mirage etc entries with documentation and `sourceUrls`
- Upgrade model to claude-opus-4-7
- Reframe Type of compliance taxonomy in schema.ts (remove KYT; add Programmatic policies / KYC/KYB / POI/ASP). **This aligns our taxonomy with Predicate and INCO report**
- Add minimal README.md outlining the research flow
- Add .gitignore (ignore *.json and other local sources)

Skills were generated by claude and are added in next PR

### DIFF
--- a/project-evaluations/scripts/research-protocol.ts
+++ b/project-evaluations/scripts/research-protocol.ts
@@ -1,8 +1,9 @@
 import "dotenv/config";
 import Anthropic from "@anthropic-ai/sdk";
+import { existsSync, readFileSync } from "fs";
 import { PROPERTY_DEFINITIONS } from "../src/data/schema.js";
 import type { Property, PropertyContent } from "../src/types.js";
-import { buildResearchPrompt, buildEvaluationPrompt } from "./research-prompts.js";
+import { buildEvaluationPrompt } from "./research-prompts.js";
 import { type ProtocolConfig } from "./research-config.js";
 
 const SKIP_PROPERTIES = new Set([
@@ -12,9 +13,37 @@ const SKIP_PROPERTIES = new Set([
   "On-chain gas cost: withdraw",
 ]);
 
-/** Loop through property definitions, skip/keep as needed, and evaluate each. */
+type CacheEntry = { name: string; url: string; summary: string };
+type ResearchCache = { id: string; generatedAt: string; properties: CacheEntry[] };
+
+export function loadResearchCache(protocolId: string): ResearchCache {
+  const cachePath = new URL(`./research-cache/${protocolId}.json`, import.meta.url);
+  if (!existsSync(cachePath)) {
+    throw new Error(
+      `Research cache missing at ${cachePath.pathname}.\n` +
+        `Run /research-sources ${protocolId} in Claude Code to populate it before invoking this script.`,
+    );
+  }
+  return JSON.parse(readFileSync(cachePath, "utf-8"));
+}
+
 export async function evaluateProperties(config: ProtocolConfig, existingProperties: Property[], only: string[]) {
   const client = new Anthropic();
+  const cache = loadResearchCache(config.id);
+  const cacheByName = new Map(cache.properties.map((entry) => [entry.name, entry]));
+
+  const targetedUrls = cache.properties
+    .filter((entry) => only.length === 0 || only.includes(entry.name))
+    .map((entry) => entry.url);
+  const uniqueUrls = Array.from(new Set(targetedUrls));
+  const fetchedByUrl = new Map<string, { text: string; sourceUrl: string }>();
+  await Promise.all(
+    uniqueUrls.map(async (url) => {
+      const fetched = await fetchTextFromUrl(url);
+      if (fetched) fetchedByUrl.set(url, fetched);
+      else console.warn(`warn: failed to fetch ${url} — properties citing it will be INSUFFICIENT_DATA`);
+    }),
+  );
 
   const properties: Property[] = [];
   let okCount = 0;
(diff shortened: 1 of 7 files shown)


project-evaluations/scripts/research-protocol.ts
```diff
@@ -115,23 +134,39 @@ async function evaluateWithCitations(
...
     ],
   });
 
-  // Extract text and deduplicated char_location citations
+  const usage = response.usage;
+  console.log(
+    `cache: read=${usage.cache_read_input_tokens ?? 0} write=${usage.cache_creation_input_tokens ?? 0} miss=${usage.input_tokens}`,
+  );
+
+  if (process.env.DEBUG_RESEARCH) {
```
NicoSerranoP: Should we add this `DEBUG_RESEARCH` env variable to the example file? I know it is an internal tool but that way we know what options we have (normal, debug, etc)

project-evaluations/scripts/research-protocol.ts
```diff
@@ -115,23 +134,39 @@ async function evaluateWithCitations(
...
+    console.log("=== end raw response.content ===");
+  }
+
   const textParts: string[] = [];
   const citations: NonNullable<Property["citations"]> = [];
   const seen = new Set<string>();
+  let toolInput: { value?: string; insufficient_data: boolean } | null = null;
 
   for (const block of response.content) {
+    if (block.type === "tool_use" && block.name === "record_evaluation") {
```
NicoSerranoP, line 167: what is the difference between `"tool_use"` and `"record_evaluation"`?

project-evaluations/src/data/schema.ts
```diff
@@ -60,8 +60,7 @@ export const PROPERTY_DEFINITIONS: PropertyContent[] = [
   {
     name: "Asset privacy",
     group: "Privacy",
-    description:
```
NicoSerranoP, line 63: Why are we deleting the `N/A` explainers around the different properties?

project-evaluations/src/data/schema.ts
```diff
@@ -234,14 +228,13 @@ export const PROPERTY_DEFINITIONS: PropertyContent[] = [
...
+    metric: "POI/ASP / Selective disclosure / KYC/KYB / Programmatic policies / Other / None",
     inputType: "multi-select",
     options: [
       "Proof of innocence (POI) / ASP",
       "Selective disclosure",
       "KYC/KYB",
-      "KYT",
```
NicoSerranoP, line 244: Why do we remove KYT? 😊

NicoSerranoP (changes requested):
Looks good! The only priority question I have is why are we reducing property definitions and in some cases deleting property options?

I can see some reasons in this PR: https://github.com/privacy-ethereum/private-transfers-benchmarks/pull/165 . My main concern was KYT because it is used for compliance:

- **KYC**: know your user information (name, phone, nationality. etc.)
- **POI:** your address is not in a sanctioned list
- **KYT:** your address has not interacted with sanctioned/malicious addresses?

project-evaluations/src/data/schema.ts
JohnGuilding: There were two sources of truth for rules: 1) in the schema 2) in the rules. This was made as an effort to keep the description as the description, and the rules as the rules.

Before, we mixed what the property is, with how to apply it to certain cases. There are still points where this is inconsistent, but this was a start to clean things up.

project-evaluations/scripts/research-protocol.ts
JohnGuilding: `record_evaluation`: the name of the custom tool the code defines. After answering, Claude calls this tool.                                                                                                                                                               

`tool_use`: We're checking if a tool is being used. when the model calls a tool, the response contains a block with `type: "tool_use"`.                            
                                                                                                                                                                              
This block checks that the block has a tool invocation and that that tool is our custom `record_evaluation`.

project-evaluations/scripts/research-protocol.ts
JohnGuilding: It was added by Claude. I will remove it

project-evaluations/src/data/schema.ts
JohnGuilding: Been thinking about this some more. Original rational was because it is not included in the INCO <> Predicate report. I was viewing it as being implicitly covered by other definitions. I'm know thinking that we just need to define it better.

> KYT: your address has not interacted with sanctioned/malicious addresses?

This is POI, not "know your transaction"?

From report (initial description and description on Programmatic Policies):

> # Compliance Solution

> There are various methods for achieving compliance, with the primary approach being programmatic policies, KYC/KYB, and inclusion/exclusion-based proofs. However, there are also further methods, such as UTXO lineage, which are detailed in the relevant case studies sections.

> ## Programmatic Policies

> A policy is a programmable set of rules that determines whether a transaction is approved or denied. In Web2 and traditional financial systems, such as banks, payment processors, and technology platforms, policies are invisible infrastructure that businesses define and use to enforce rules based on their specific requirements. Similarly, policies can be used by Web3 developers to enforce business, legal, ethical, and regulatory requirements for privacy applications built on blockchains.

> In this approach, first developed by Predicate, privacy application developers define a set of pre-transaction rules, which are combined to form policies. The developer determines the categories of high-risk addresses that they want blocked, such as addresses involved with sanctions, terrorist financing, human trafficking, fraud, etc. Then, they determine the onchain and offchain information sources used to identify if an address seeking to deposit funds into the privacy application is associated with any high-risk activity, such as:

> - Blockchain analytics firms’ risk scores (e.g. Chainalysis, TRM, Elliptic)
> - Information Sharing and Analysis Center (ISAC) data (e.g. SEAL-Intel)
> - Fraud detection sources (e.g. Webacy)
> - Offchain KYC/KYB platforms (e.g. Sumsub, AiPrise, Persona)
> - ZK identity tools (e.g. ZKPassport, Self, Onflow)
> - Historical transaction volume

Working and expanding definition of KYT:

> KYT (Know Your Transaction): compliance mechanism where the subject is the transaction itself. KYT screening includes assessing counterparty address risk (SIMILAR TO POI — NEEDS CLEANING UP), fund flow, amounts, or behavioural patterns. As opposed to the user's identity (KYC/KYB) or cryptographic set membership (POI/ASP). Can be implemented as a programmatic policy, or as an independent policy

I've added [an issue](https://github.com/privacy-ethereum/private-transfers-benchmarks/issues/178) to look at re-adding it.

project-evaluations/scripts/research-protocol.ts
JohnGuilding: The logic is complex, I have added some comments to aid understanding. Open to simplifying actual logic in future. I'm a bit hesitant rn as the logic is quite brittle and can change the quality of the analysis a lot when making changes. Would rather do a full review pass of all currently raised PRs before messing with logic again

NicoSerranoP (approved):

PR #170, opened 2026-04-21
Feat/add mirage analysis

### LINKED ISSUE #68: Mirage
### Requirements:
* All non-benchmark assessment criteria filled in
    * The writing should be the same quality and detail as the property analysis from the State of Private Voting Report 2026
* If using AI-assisted research:
    * References links from docs/code/whitepapers must exist
    * AI filler, slop and hallucinations should be removed

### Out of scope properties:
1. Deposit benchmarks
2. Transfer benchmarks
3. Withdraw benchmarks
4. Anonymity set size

JohnGuilding:
Add mirage analysis

### DIFF
--- a/project-evaluations/src/data/evaluations/mirage.json
+++ b/project-evaluations/src/data/evaluations/mirage.json
@@ -1,0 +1,557 @@
+{
+  "id": "mirage",
+  "title": "Mirage",
+  "description": "Mirage is a privacy layer for stablecoin transfers on Ethereum L1 that makes private transactions indistinguishable from ordinary onchain activity. It avoids shared pools by using per-transaction escrow contracts whose EVM bytecode is obfuscated by Azoth, an open-source Rust obfuscator that produces many deterministic variants of the same contract logic. Execution is carried out by a decentralised set of Nomad operator nodes that post bonds into the escrow and earn rewards for settling transfers. The protocol is currently in closed alpha with a public test environment in preparation.",
+  "documentation": "https://docs.mirageprivacy.com/",
+  "categories": ["EVM Obfuscators"],
+  "properties": [
+    {
+      "name": "Anonymity set size",
+      "value": ""
+    },
+    {
+      "name": "Confidentiality",
+      "value": "No",
+      "notes": "Amounts and balances are public, so the answer is No. Alice's wallet deposits the payout (100 USDC) plus a 2 USDC tip into the per-transaction escrow , and the node then sends 100 USDC from its own wallet directly to Bob, so on-chain it looks like a random address transferring 100 USDC . Transfer amounts and per-address balances of the underlying stablecoin are therefore visible on-chain in plaintext; the privacy property provided is unlinkability between sender and recipient, not confidentiality of amounts.",
+      "citations": [
+        {
+          "cited_text": "Alice’s wallet deposits into this escrow:\n\n100 USDC (the amount she wants Bob to receive)\n\n2 USDC (a small tip to reward whoever completes the transfer)\n\nAt the same time, the Mirage app encrypts Alice’s instructions: “Send 100 USDC to Bob’s address. ",
+          "source": "https://docs.mirageprivacy.com/understanding-mirage/a-simple-example",
+          "start_char_index": 1480,
+          "end_char_index": 1731
+        },
+        {
+          "cited_text": "If they want to take the job, here&#x27;s what happens:\n\nThe node posts a security deposit into Alice&#x27;s escrow contract (ensuring they only profit by completing the job honestly)\n\nThe node sends 100 USDC from their own wallet directly to Bob via another address \n\nThis is the key moment. Onchain, it looks like this node (some random address) is sending Bob 100 USDC. ",
+          "source": "https://docs.mirageprivacy.com/understanding-mirage/a-simple-example",
+          "start_char_index": 2326,
+          "end_char_index": 2699
+        }
+      ]
+    },
+    {
+      "name": "Anonymity",
+      "value": "Unlinkability",
+      "notes": "Mirage provides unlinkability rather than hidden endpoints: both the sender's escrow deployment and the recipient's incoming transfer are visible on-chain, but nothing obvious ties them together. Instead of a direct sender-to-recipient transaction, the app sets up a mechanism where a third party sends the funds on the sender's behalf, so to a blockchain observer it appears that a random address sent the recipient the amount. A node posts a security deposit into the escrow and then sends the funds from its own wallet to the recipient via another address; on-chain this looks like a random address paying the recipient, with no visible connection back to the sender. An observer sees the sender deploying a contract and depositing funds, a separate random address later paying the recipient, and another address withdrawing from the contract — three events with no visible connection between sender and recipient. The sender's funding transaction and the recipient's receipt are both publicly observable as individual events, so neither party is hidden in isolation; the privacy property achieved is the breaking of the linkage between them via the intermediary escrow plus node settlement.",
+      "citations": [
+        {
+          "cited_text": "Instead of Alice sending directly to Bob (which would be visible onchain and link them together), the Mirage app sets up a mechanism where someone else sends the money to Bob on Alice’s behalf. To anyone watching the blockchain, it just looks like a random address sent Bob 100 USDC. ",
+          "source": "https://docs.mirageprivacy.com/understanding-mirage/a-simple-example",
+          "start_char_index": 662,
+          "end_char_index": 946
+        },
+        {
+          "cited_text": "If they want to take the job, here&#x27;s what happens:\n\nThe node posts a security deposit into Alice&#x27;s escrow contract (ensuring they only profit by completing the job honestly)\n\nThe node sends 100 USDC from their own wallet directly to Bob via another address \n\nThis is the key moment. Onchain, it looks like this node (some random address) is sending Bob 100 USDC. There&#x27;s no visible connection to Alice. ",
+          "source": "https://docs.mirageprivacy.com/understanding-mirage/a-simple-example",
+          "start_char_index": 2326,
+          "end_char_index": 2744
+        },
+        {
+          "cited_text": "Why This Provides Privacy \n\nLet&#x27;s step back and look at what an observer sees when analyzing the blockchain:\n\nTransaction 1: Alice deployed a contract and deposited some USDC into it\nInterpretation: Could be anything. A multisig, an MEV contract, an automated trading bot, whatever \n\nTransaction 2: Some random address sent Bob 100 USDC\nInterpretation: Just a normal transfer. Happens thousands of times per day \n\nTransaction 3: Another random address later withdrew funds from Alice&#x27;s contract\nInterpretation: who knows? \n\nThere&#x27;s no visible connection between Alice and Bob. ",
+          "source": "https://docs.mirageprivacy.com/understanding-mirage/a-simple-example",
+          "start_char_index": 3659,
+          "end_char_index": 4251
+        }
+      ]
+    },
+    {
+      "name": "Asset privacy",
+      "value": "No",
+      "notes": "No. Mirage is a stablecoin transfer protocol where the asset being moved is directly observable on-chain — the blockchain activity looks like an ordinary stablecoin transfer, with observers able to see a deposit into a temporary contract, a standard transfer from another address to the recipient, and a later withdrawal from the escrow contract . Nodes provide their own stablecoin liquidity (e.g. USDC) to execute transfers , and the token identity is visible at both escrow deposit and recipient payout with no commitment scheme concealing which asset moves. Privacy is scoped to unlinking sender and recipient rather than hiding asset identity.",
+      "citations": [
+        {
+          "cited_text": "Mirage is designed so that the blockchain activity looks like an ordinary stablecoin transfer.\n\nObservers may see:\n\nA deposit into a temporary contract\n\nA standard transfer from another address to the recipient\n\nA later withdrawal from the escrow contract\n\nBecause each transaction uses a unique contract and does not interact with a recognizable privacy pool, observers cannot determine that a privacy protocol was used without spending significant resources on identifying a specific target.\n\n",
+          "source": "https://docs.mirageprivacy.com/faq/general",
+          "start_char_index": 4813,
+          "end_char_index": 5308
+        },
+        {
+          "cited_text": "Mirage transactions rely on nodes that provide their own liquidity to execute transfers. Nodes must maintain a minimum amount of stablecoin liquidity (e.g. USDC) in order to execute user requests.\n\n",
+          "source": "https://docs.mirageprivacy.com/faq/general",
+          "start_char_index": 5914,
+          "end_char_index": 6112
+        }
+      ]
+    },
+    {
+      "name": "Plausible deniability",
+      "value": "Yes",
+      "notes": "Entry into the protocol is designed to be indistinguishable from an ordinary on-chain transfer. A stated core idea is that no observer should know a privacy protocol was used, positioning this as a design goal differentiating it from existing privacy solutions which are easy to flag when users interact with CEXs or on/off-ramps. In practice, per-transaction escrow bytecode is rewritten into deterministic variants so the deposit resembles a routine transfer to an unverified contract, and the recipient sees a USDC transfer from a fresh EOA, leaving no \"Mirage\"-tagged footprint on either side. Note: deniability applies to the deposit/entry path; the settlement leg performed by the Nomad network is the visible protocol-side interaction and is not covered by sender-side deniability. The protocol is in closed alpha as of April 2026, so this assessment reflects the documented design rather than mainnet behaviour.",
+      "needsResearchReview": true,
+      "citations": [
+        {
+          "cited_text": "Menu\nChevron Down \n\nMenu \n\nThe Undetectability Challenge\n\nOn this page\nChevron Right \n\nThe Undetectability Challenge \n\nOne of the core ideas behind Mirage is that no one should know you used a privacy protocol to begin with .\n\nMirage is the first privacy solution that makes an attempt at protecting its users against the high-risk stigma of using privacy protocols. ",
+          "source": "https://docs.mirageprivacy.com/understanding-mirage/the-undetectability-challenge",
+          "start_char_index": 168,
+          "end_char_index": 535
+        }
+      ]
+    },
+    {
+      "name": "On-chain gas cost: deposit",
+      "value": ""
+    },
+    {
+      "name": "On-chain gas cost: transfer",
+      "value": ""
+    },
+    {
+      "name": "On-chain gas cost: withdraw",
+      "value": ""
+    },
+    {
+      "name": "Time-to-finality",
+      "value": "N/A",
+      "notes": "Time-to-finality is N/A because Mirage is an application-layer protocol that settles on host chains and inherits their finality. End-to-end processing times quoted are under ~90 seconds on Ethereum mainnet, under ~6 seconds on Layer 2 networks such as Base or Arbitrum, and under ~4 seconds on the Tempo testnet, though actual processing time depends on network block times and congestion. These represent node settlement latency rather than cryptographic finality, which is determined by the underlying chain's consensus.",
+      "citations": [
+        {
+          "cited_text": "The following time indications represent the time for the network of nodes to process and settle the transaction:\n\nEthereum mainnet: less than ~90 seconds\n\nLayer 2 networks (e.g. Base or Arbitrum): under ~6 seconds\n\nTempo (Testnet): under ~4 seconds\n\nActual processing time depends on network block times and congestion.\n\n",
+          "source": "https://docs.mirageprivacy.com/faq/general",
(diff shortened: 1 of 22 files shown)


project-evaluations/src/data/evaluations/mirage.json
```diff
@@ -1,10 +1,595 @@
 {
   "id": "mirage",
   "title": "Mirage",
-  "description": "",
-  "status": "pending",
-  "documentation": "",
+  "description": "Mirage is a privacy layer for stablecoin transfers on Ethereum that makes private transactions indistinguishable from ordinary onchain activity. It does this via a technique called bytecode obfuscation along with some other components. It avoids shared pools by using per-transaction escrow contracts whose EVM bytecode is obfuscated by Azoth, an open-source Rust obfuscator that produces many deterministic variants of the same contract logic. Execution is carried out by a decentralised set of TEE-based nodes called Nomads that post bonds into the escrow and earn rewards for settling transfers.",
```
NicoSerranoP:
```suggestion
  "description": "Mirage is a privacy layer for token transfers on Ethereum that makes private transactions indistinguishable from ordinary onchain activity. It does this via a technique called bytecode obfuscation along with some other components. It avoids shared pools by using per-transaction escrow contracts whose EVM bytecode is obfuscated by Azoth, an open-source Rust obfuscator that produces many deterministic variants of the same contract logic. Execution is carried out by a decentralised set of TEE-based nodes called Nomads that post bonds into the escrow and earn rewards for settling/finalizing transfers.",
```

project-evaluations/src/data/evaluations/mirage.json
```diff
@@ -1,10 +1,595 @@
...
-  "properties": [],
-  "sourceUrls": []
+  "properties": [
+    {
+      "name": "Anonymity set size",
+      "value": ""
+    },
+    {
+      "name": "Confidentiality",
+      "value": "No",
```
NicoSerranoP, line 14:
```suggestion
      "value": "Only balances",
```

NicoSerranoP: See comment below

project-evaluations/src/data/evaluations/mirage.json
```diff
@@ -1,10 +1,595 @@
...
+      "notes": "Transfer amounts and per-address balances of the underlying stablecoin are visible on-chain in plaintext, with no shielded commitments or encrypted amounts. Alice's wallet deposits the payout (100 USDC) plus a 2 USDC tip into the per-transaction escrow, and the node then sends 100 USDC from its own wallet directly to Bob, so on-chain it looks like a random address transferring 100 USDC.",
```
NicoSerranoP:
```suggestion
      "notes": "The sender deploys a escrow smart contract and sends funds to it. The transfer amounts are visible on-chain in plaintext, with no shielded commitments or encrypted amounts. There is no aggregated private balance because each transaction requires a newly deployed escrow contract.  After the escrow deployment, the Nomad network (TEEs) take the transfer request and send the transfer amount to the receiver. Hence we consider that Mirage only provides confidentiality in the users' balances and not in the transfer amounts",
```

project-evaluations/src/data/evaluations/mirage.json
NicoSerranoP: Mirage testnet app has USDC and ETH available for private transfers. That is why I removed the "stablecoin" term

project-evaluations/src/data/evaluations/mirage.json
```diff
@@ -1,10 +1,595 @@
...
+      "value": ""
+    },
+    {
+      "name": "On-chain gas cost: withdraw",
+      "value": ""
+    },
+    {
+      "name": "Time-to-finality",
+      "value": "N/A",
+      "notes": "Mirage is an application-layer protocol that settles on host chains and inherits their finality guarantees, so finality is determined by the underlying chain's consensus rather than by anything Mirage adds. The design targets Ethereum, with adaptations for L2s such as Arbitrum and Base, and other L1s like Solana. For maximum privacy, transfers are multiple transactions and take place over several blocks — under ~90 seconds on Ethereum mainnet, under ~6 seconds on L2s, and under ~4 seconds on the Tempo testnet — describe node-settlement latency rather than cryptographic finality.",
```
NicoSerranoP:
```suggestion
      "notes": "Mirage is an application-layer protocol that settles on host chains and inherits their finality guarantees, so finality is determined by the underlying chain's consensus rather than by anything Mirage adds. The design targets Ethereum, with adaptations for L2s such as Arbitrum and Base, and other L1s like Solana. For maximum privacy, transfers are multiple transactions and take place over several blocks — under ~90 seconds on Ethereum mainnet, under ~6 seconds on L2s, and under ~4 seconds on the Tempo testnet — describe node-settlement latency rather than cryptographic finality. A Mirage private transfer needs to be executed by a node of the Nomad network (TEEs) that is constantly listening on-chain for encrypted private transfer requests",
```

project-evaluations/src/data/evaluations/mirage.json
```diff
@@ -1,10 +1,595 @@
...
+          "source": "https://docs.mirageprivacy.com/faq/general",
+          "start_char_index": 4116,
+          "end_char_index": 4438
+        }
+      ]
+    },
+    {
+      "name": "Number of secrets",
+      "value": "1",
+      "notes": "Mirage requires the user to hold only their existing Ethereum wallet key (secp256k1 ECDSA), which authorises deployment of and deposit into the per-transaction escrow. The protocol derives no viewing key, spending key, or nullifier seeds. The recipient is paid at their normal address with no Mirage-specific signing. Wallets broadcast an encrypted message called a signal — signal encryption is handled by the client and network rather than from a user-held long-term key.",
```
NicoSerranoP:
```suggestion
      "notes": "Mirage requires the user to hold only their existing Ethereum wallet key (secp256k1 ECDSA), which authorises deployment of and deposit into the per-transaction escrow. The protocol derives no viewing key, spending key, or nullifier seeds. The recipient is paid at their normal address with no Mirage-specific signing. Wallets broadcast an encrypted message called a signal — signal encryption is handled by the client and network rather than from a user-held long-term key. The obfuscated escrow contract can be deterministically verified by using public information (like transaction parameters) as seed. No secrets are required to perform a verification of a Mirage contract.",
```

project-evaluations/src/data/evaluations/mirage.json
```diff
@@ -1,10 +1,595 @@
...
+          "source": "https://docs.mirageprivacy.com/understanding-mirage/a-simple-example",
+          "start_char_index": 1480,
+          "end_char_index": 1731
+        },
+        {
+          "cited_text": "When you want to make a private transaction, your wallet broadcasts an encrypted message (called a &#x27;signal&#x27;) to a network of independent executors. Each signal is locked behind a cryptographic puzzle that must be solved before its contents can be read. Your actual intentions are only revealed to the node that picks up your transaction. ",
+          "source": "https://docs.mirageprivacy.com/understanding-mirage/introduction",
+          "start_char_index": 3439,
+          "end_char_index": 3787
+        }
```
NicoSerranoP, line 151:
```suggestion
        },
        {
            "cited_text": "The seed can be derived deterministically from public information (like the transaction parameters), so verification doesn't require any secrets. Anyone can check any Mirage contract at any time.",
            "source": "https://docs.mirageprivacy.com/azoth/introduction/#why-determinism-matters"
        }
```

project-evaluations/src/data/evaluations/mirage.json
```diff
@@ -1,10 +1,595 @@
...
+          "cited_text": "Alice’s wallet deposits into this escrow:\n\n100 USDC (the amount she wants Bob to receive)\n\n2 USDC (a small tip to reward whoever completes the transfer)\n\nAt the same time, the Mirage app encrypts Alice’s instructions: “Send 100 USDC to Bob’s address. ",
+          "source": "https://docs.mirageprivacy.com/understanding-mirage/a-simple-example",
+          "start_char_index": 1480,
+          "end_char_index": 1731
+        }
+      ]
+    },
+    {
+      "name": "Withdraw time",
+      "value": "0",
```
NicoSerranoP, line 175:
```suggestion
      "value": "90",
```

NicoSerranoP: See comment below

project-evaluations/src/data/evaluations/mirage.json
```diff
@@ -1,10 +1,595 @@
...
+      "notes": "Withdrawal from the per-transfer escrow is available immediately when no Nomad has sent a bond: funds can be withdrawn from the smart contract at any time without permissions, as long as a node has not already submitted a security deposit. Once a node sends it's bond, withdrawal is paused until either settlement completes or the bond deadline lapses — Nomads post bonds of at least half the reward to secure execution rights within a 5-minute deadline, after which forfeited bonds roll into the reward pool and the deployer can withdraw.",
```
NicoSerranoP, line 176:
```suggestion
      "notes": "The withdrawal process in Mirage starts as soon as a node of the Nomad network sends a bond and start the payment process. The node has to send the transfer amount to the recipient from his own wallet and prove that the recipient got the funds to receive the funds in the escrow contract (transfer amount + processing fee). The process can take 90 seconds according to the protocol documentation.",
```

project-evaluations/src/data/evaluations/mirage.json
```diff
@@ -1,10 +1,595 @@
...
+          "source": "https://docs.mirageprivacy.com/faq/general",
+          "start_char_index": 4116,
+          "end_char_index": 4438
+        }
+      ]
+    },
+    {
+      "name": "Censorship resistance",
+      "value": "No",
+      "notes": "There are several ways in which a transfer could be stopped. The platform can revoke the enclave binary that all Nomad operators run, blocking clients from producing usable signals. Onboarding as a Nomad is currently handled directly by the team rather than open registration.",
```
NicoSerranoP:
```suggestion
      "notes": "There are several ways in which a transfer could be stopped. The platform can revoke the enclave binary that all Nomad operators run, blocking clients from producing usable signals. Onboarding as a Nomad is currently handled directly by the team rather than open registration. Mirage runs a wallet screening check through a third-party compliance partner and the Mirage API could stop issuing compliance certificates to block certain users from participating.",
```

project-evaluations/src/data/evaluations/mirage.json
```diff
@@ -1,10 +1,595 @@
...
+          "source": "https://docs.mirageprivacy.com/faq/general",
+          "start_char_index": 3870,
+          "end_char_index": 4058
+        }
+      ]
+    },
+    {
+      "name": "Upgradeability",
+      "value": "Single admin",
+      "notes": "Upgrade authority sits with the Mirage team. Although each deployed escrow instance is immutable, the team controls the toolchain and operator set that define protocol behaviour across future transfers. The Azoth bytecode obfuscator, and any upcoming escrow contract upgrades are all team-controlled. Instance-level immutability does not constrain these upgrades. However, both the escrow contract and Azoth are open source. Azoth's outputs are also deterministic, meaning it is possible to perform an integrity check to ensure the contract you're interacting with hasn't been tampered with.",
```
NicoSerranoP, line 258:
```suggestion
      "notes": "Upgrade authority sits with the Mirage team. Although each deployed escrow instance is immutable, the team controls the toolchain and operator set that define protocol behaviour across future transfers. Instance-level immutability does not constrain these upgrades. However, both the escrow contract and Azoth are open source. Azoth's outputs are also deterministic, meaning it is possible to perform an integrity check to ensure the contract you're interacting with hasn't been tampered with.",
```

NicoSerranoP: I would remove that sentence because Azoth is open source under the MIT License. Changes to the obfuscator would not affect the running protocol with an existing version of Azoth

project-evaluations/src/data/evaluations/mirage.json
```diff
@@ -1,10 +1,595 @@
...
+          "source": "https://docs.mirageprivacy.com/faq/compliance",
+          "start_char_index": 497,
+          "end_char_index": 651
+        }
+      ]
+    },
+    {
+      "name": "Type of compliance",
+      "value": "[\"Programmatic policies\"]",
+      "notes": "Compliance is enforced at escrow deployment via a wallet-screening check run by a third-party compliance partner — wallet screening, risk scoring, and analysis of the origin of funds. The check happens at the Mirage web application layer rather than on-chain.",
```
NicoSerranoP:
```suggestion
      "notes": "Compliance is enforced at escrow deployment via a wallet-screening check run by a third-party compliance partner — wallet screening, risk scoring, and analysis of the origin of funds. The check happens at the Mirage web application layer rather than on-chain. The API logic can be modified to work with new policies",
```

project-evaluations/src/data/evaluations/mirage.json
```diff
@@ -1,10 +1,595 @@
...
+          "source": "https://docs.mirageprivacy.com/understanding-mirage/introduction",
+          "start_char_index": 3382,
+          "end_char_index": 3597
+        }
+      ]
+    },
+    {
+      "name": "Client-side indexing",
+      "value": "No scanning",
+      "notes": "No scanning is required on the client side. Recipients receive funds through a direct on-chain USDC transfer from a node created wallet, which appears as a regular address sending them stablecoins, so their wallet picks it up as any ordinary incoming transfer.",
```
NicoSerranoP:
```suggestion
      "notes": "No scanning is required on the client side. Recipients receive funds through a direct on-chain token transfer from a node created wallet, which appears as a regular address sending them funds, so their wallet picks it up as any ordinary incoming transfer.",
```

project-evaluations/src/data/evaluations/mirage.json
```diff
@@ -1,10 +1,595 @@
...
+          "source": "https://docs.mirageprivacy.com/nomad/overview/",
+          "start_char_index": 737,
+          "end_char_index": 1309
+        }
+      ]
+    },
+    {
+      "name": "Access to DeFi",
+      "value": "No access to DeFi",
+      "notes": "Mirage operates as a stablecoin transfer layer that enables confidential stablecoin transfers directly on public blockchains, with no protocol token, shielded asset, or wrapper that could be composed into DeFi. The on-chain footprint consists of a deposit into a temporary contract, a standard transfer from another address to the recipient, and a later withdrawal from the escrow.",
```
NicoSerranoP:
```suggestion
      "notes": "Mirage operates as a token transfer layer that enables confidential transfers directly on public blockchains, with no protocol token, shielded asset, or wrapper that could be composed into DeFi. The on-chain footprint consists of a deposit into a temporary contract, a standard transfer from another address to the recipient, and a later withdrawal from the escrow.",
```

project-evaluations/src/data/evaluations/mirage.json
```diff
@@ -1,10 +1,595 @@
...
+          "source": "https://docs.mirageprivacy.com/understanding-mirage/introduction",
+          "start_char_index": 2368,
+          "end_char_index": 2514
+        }
+      ]
+    },
+    {
+      "name": "Programmability / Generality",
+      "value": "Only payments",
+      "notes": "Mirage's programmability is limited to payment flows, specifically private stablecoin transfers executed through a one-time escrow pattern. Each transaction deploys a unique smart contract that to outside observers looks like a normal, unverified contract.",
```
NicoSerranoP:
```suggestion
      "notes": "Mirage's programmability is limited to payment flows, specifically private token transfers executed through a one-time escrow pattern. Each transaction deploys a unique smart contract that to outside observers looks like a normal, unverified contract.",
```

NicoSerranoP (approved):
I left some important comments and modifications 😊 . Mostly related to changes from stablecoins to tokens because Mirage allows USDC and ETH at the moment. 

The escrow contract logic is generic for any ERC20 token. The Nomad network requires enough ERC20 token liquidity to keep the protocol working

project-evaluations/src/data/evaluations/mirage.json
JohnGuilding: Are you certain? Would the user just be moved to the new escrow contracts without noticing if an update was pushed?

project-evaluations/src/data/evaluations/mirage.json
JohnGuilding: So for confidentiality we have privacy of balances and privacy of transfer amounts.

The users balance is still viewable since it is held in their regular Ethereum account, same with the recipient. the escrow contracts are not balances, they are transfer amounts. A transfer is normally a single value that can be analysed in isolation i.e. there is no notion of an aggregated transfer amount by default. There is just a transfer amount. In this case, the transfer amount is visible at the senders side, and at the recipients side. So based on that information, this would still be a no

A separate point is based on previous analysis of protocols that provided some unlinkability mechanism, we have marked them all as "no" for confidentiality. So would have to change other protocols if making this change (which is fine and is what we should do if we notice an error)

Wdyt? Do you disagree with any of these assumptions or arguments?

Will remove "stablecoin"

project-evaluations/src/data/evaluations/mirage.json
JohnGuilding: This is based on the underlying chain rather than the protocol from my understanding? It is 90 seconds for Ethereum but only a few seconds for L2s

https://docs.mirageprivacy.com/faq/general#how-long-does-it-take-to-generate-private-transactions

So far, we've based analysis on all deployments rather than mainnet specific deployments if multiple exist

project-evaluations/src/data/evaluations/mirage.json
NicoSerranoP: Lets keep confidentiality as "No"

project-evaluations/src/data/evaluations/mirage.json
NicoSerranoP: It is based on the underlying chain + the Nomad node decrypting the request + Nomad node sending the transfer amount to the recipient.

This whole process will take more than a L1 time-finality because the Nomad node needs to listen, decrypt, send transfer and fetch the funds in the escrow.

It is not super clear how long would it take but I would go with the 90 seconds (for Mainnet) stated in the docs

project-evaluations/src/data/evaluations/mirage.json
NicoSerranoP: Existing unclaimed escrows will remain the same and only new escrow contracts will use the new Azoth version. I feel this might not be a problem because users will be able to check the Azoth source code and verify it builds the required escrow

project-evaluations/src/data/evaluations/mirage.json
JohnGuilding: nice, changing

project-evaluations/src/data/evaluations/mirage.json
JohnGuilding: Adding

project-evaluations/src/data/evaluations/mirage.json
JohnGuilding: Adding

project-evaluations/src/data/evaluations/mirage.json
JohnGuilding: Adding

project-evaluations/src/data/evaluations/mirage.json
JohnGuilding: Adding

project-evaluations/src/data/evaluations/mirage.json
JohnGuilding: Adding

project-evaluations/src/data/evaluations/mirage.json
JohnGuilding: Adding

project-evaluations/src/data/evaluations/mirage.json
JohnGuilding: Adding

project-evaluations/src/data/evaluations/mirage.json
JohnGuilding: Nice adding

project-evaluations/src/data/evaluations/mirage.json
JohnGuilding: Escrows are not re-used though, and they are meant to be used straight away from my understanding. So as soon the obfuscator is updated, any users who want to use the protocol from that point forwards, will have to use this upgrade - so the sentence reflects a real upgrade path.

Regarding checking the source code, I'm unsure if that changes things. That is the same with many upgrades.

project-evaluations/src/data/evaluations/mirage.json
JohnGuilding: 90 seconds would be misleading as different chains have different values according to docs. We would need to explicitly say this is mainnet only, which we haven't done with the analysis so far — all properties are chain-neutral.

We would have to change this assumption, we can create an issue to revisit this? 

I'm also unsure if this a withdraw time, this is more transfer time

project-evaluations/src/data/evaluations/mirage.json
JohnGuilding: Let me know if you still disagree, but this feels like a real upgrade path to me

PR #172, opened 2026-04-21
feat: subgraph client requests

0xmad:
- [x] Add code generation for graphql types
- [x] Add boilerplate for project stats queries
- [x] Add SUBGRAPH_URL env

### DIFF
--- a/.github/workflows/root.yml
+++ b/.github/workflows/root.yml
@@ -28,6 +28,8 @@ jobs:
 
       - name: Lint
         run: pnpm run lint
+        env:
+          SUBGRAPH_URL: ${{ secrets.SUBGRAPH_URL }}
 
       - name: Format
         run: pnpm run format
--- a/gas-benchmarks/.env.example
+++ b/gas-benchmarks/.env.example
@@ -4,4 +4,6 @@ SCROLL_RPC_URL=https://scroll.gateway.tenderly.co
 
 MONERO_FAIL_NODES_API_URL=https://monero.fail/nodes.json
 
+SUBGRAPH_URL=https://api.studio.thegraph.com/query/<USER_ID>/<SUBGRAPH_SLUG>/<VERSION>
+
 MIN_SAMPLES=10
--- a/gas-benchmarks/codegen.ts
+++ b/gas-benchmarks/codegen.ts
@@ -1,0 +1,28 @@
+import { config } from "dotenv";
+import path from "node:path";
+
+import type { CodegenConfig } from "@graphql-codegen/cli";
+
+config();
+
+if (!process.env.SUBGRAPH_URL) {
+  throw new Error("SUBGRAPH_URL is not set");
+}
+
+const codegenConfig: CodegenConfig = {
+  overwrite: true,
+  schema: [process.env.SUBGRAPH_URL, path.resolve(".", "scalars.graphql")],
+  ignoreNoDocuments: true,
+  documents: ["src/**/*.ts"],
+  emitLegacyCommonJSImports: false,
+  generates: {
+    "src/generated/": {
+      preset: "client-preset",
+      config: {
+        useTypeImports: true,
+      },
+    },
+  },
+};
+
+export default codegenConfig;
--- a/gas-benchmarks/scalars.graphql
+++ b/gas-benchmarks/scalars.graphql
@@ -1,0 +1,3 @@
+scalar Bytes
+scalar BigInt
+scalar BigDecimal
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "private": true,
   "type": "module",
   "scripts": {
+    "prelint": "pnpm -r run codegen",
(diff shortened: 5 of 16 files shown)


.github/workflows/root.yml
```diff
@@ -28,6 +28,8 @@ jobs:
 
       - name: Lint
         run: pnpm run lint
+        env:
```
NicoSerranoP, line 31: Why do we need env variables in the root workflow? It only checks that lint and format run correctly

gas-benchmarks/.env.example
```diff
@@ -4,4 +4,6 @@ SCROLL_RPC_URL=https://scroll.gateway.tenderly.co
 
 MONERO_FAIL_NODES_API_URL=https://monero.fail/nodes.json
 
+SUBGRAPH_URL=
```
NicoSerranoP: Can we put an example like:

`SUBGRAPH_URL=https://api.studio.thegraph.com/query/<YOUR_USER_ID>/<YOUR_SUBGRAPH_SLUG>/version/latest`

gas-benchmarks/codegen.ts
```diff
@@ -0,0 +1,28 @@
+import { config } from "dotenv";
+import path from "node:path";
+
+import type { CodegenConfig } from "@graphql-codegen/cli";
+
+config();
+
+if (!process.env.SUBGRAPH_URL) {
```
NicoSerranoP, line 8: Can we import the `SUBGRAPH_URL` variable from the `src/utils/constants.ts` file so we delete env variable configuration in this file? I am referring to:

```typescript
import { config } from "dotenv";
// ...

config();

if (!process.env.SUBGRAPH_URL) {
  throw new Error("SUBGRAPH_URL is not set");
}
```

gas-benchmarks/scalars.graphql
NicoSerranoP, line 1: Question: what does this file do? 😊 is it related to the imports in the subgraph package? For example:

```typescript
import { BigInt, Bytes } from "@graphprotocol/graph-ts";
```

.github/workflows/root.yml
NicoSerranoP: Is it because we need to run `pnpm run pre-lint` before `pnpm run lint`? 😊

package.json
```diff
@@ -5,6 +5,7 @@
   "private": true,
   "type": "module",
   "scripts": {
+    "prelint": "pnpm -r run codegen",
     "lint": "eslint gas-benchmarks/src project-evaluations/src subgraph/src subgraph/tests",
```
NicoSerranoP, line 9: What do you think about grouping these commands into one:

```json
 "scripts": {
    "prelint": "pnpm -r run codegen",
    "eslint": "eslint gas-benchmarks/src project-evaluations/src subgraph/src subgraph/tests", 
    "lint": "pnpm run prelint && pnpm run eslint",
  }
```

That way we only run `pnpm run lint` but we still leave the script declarations more readable in case someone needs to know what is happening under the hood

NicoSerranoP (approved):
Thank you Anton! Looks good! 🚀  I left some suggestions that are not high priority but could be implemented before merging the PR

.github/workflows/root.yml
0xmad: Yes, it's required to run the codegen to pass the lint check

gas-benchmarks/codegen.ts
0xmad: Not sure if this import works (it's outside src folder)

gas-benchmarks/.env.example
0xmad: Why do we need an example? It can be a different url format (especially if you use non-published version of subgraph)

gas-benchmarks/scalars.graphql
0xmad: These are scalars to support subgraph types inside graphql client

package.json
0xmad: When you run lint you can see that there is a prelint command output as well. It's a standard way to run pre/post hooks for package.json scripts.

gas-benchmarks/codegen.ts
0xmad: Plus, we need only `SUBGRAPH_URL` for codegen command but importing it from constants will require to add all of the envs for lint ci step.

gas-benchmarks/.env.example
NicoSerranoP: In order to follow the other env variables that we have on top. I am open to not put an example but I think an example could give people. Up to you 😊

package.json
NicoSerranoP: I get it, but we could have `pnpm run lint` with the prelint command inside as well. Also up to you 😊

0xmad: prelint will run with lint anyway (it will be 3 commands in total, 2 prelint and one eslint)

PR #173, opened 2026-04-22
feat(subgraph): add redact shield native

NicoSerranoP:
1. Merge multiple `PROTOCOL_NAME.graphql` files into a main `schema.graphql` so we can have shorter files per protocol. The merging process uses NodeJS to allow usage in Linux and Windows (`cat schemas/*.graphql > schema.graphql` does not work in Windows 😢 )

2. Organize handlers and ABIs into folders with the PROTOCOL name. There could be some cases where the protocol has more than 1 ABI (for example Tornado Cash or Intmax)

3.  Create 2 different subgraph.yml files for the different networks (mainnet for Railgun and sepolia for Redact). We cannot have [more than one network per subgraph.yml file ](https://thegraph.com/docs/en/subgraphs/developing/creating/subgraph-manifest/#subgraph-capabilities). I am not sure if we can [create a subgraph.yml template file like in maci-platform](https://github.com/privacy-ethereum/maci-platform/blob/ca078b796740b82dcadc18246bc637ec706e6fcf/packages/subgraph/templates/subgraph.template.yaml) due to the difference between number of events. We could leave it as a task for another PR.

4. Added Redact ShieldedNative event logic (schema, handler). I will add the Unshielded logic in an upcoming PR. I wanted to keep this one short)

5. I will add tests for ShieldedNative for Redact in the upcoming PR. I feel this current PR is large enough


subgraph/src/redact/confidential-eth.ts
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  return stats;
+}
+
+export function handleShieldedNative(event: ShieldedNativeEvent): void {
+  const id = `${event.transaction.hash.toHex()}-${event.logIndex.toString()}`;
+
+  const shield = new RedactShieldedNative(id);
+  const stats = createOrLoadProtocolStats();
```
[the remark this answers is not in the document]
NicoSerranoP: I will add tests in an upcoming PR because I do not want to make this PR bigger. For the tests I need:

1. Move railgun tests to a specific directory
2. Create redact tests

subgraph/configuration/subgraph.mainnet.yaml
```diff
@@ -0,0 +1,35 @@
+specVersion: 1.3.0
+indexerHints:
+  prune: auto
+schema:
+  file: ../schema.graphql
+dataSources:
+  - kind: ethereum
```
0xmad, line 7: Have you tried to add new item here to support multiple networks?

```
- kind: ethereum
  name: RailgunSmartWallet
  network: mainnet
  source:
    address: "0xfa7093cdd9ee6932b4eb2c9e1cde7ce00b1fa4b9"
    abi: RailgunSmartWallet
    startBlock: 14693013
   ...
- kind: ethereum
    name: ConfidentialETH
    network: sepolia
    source:
      address: "0xC132c8a82A24Fe1e491082932e3db4F70Ce95c93"
      abi: ConfidentialETH
      startBlock: 10622330
    mapping:
      kind: ethereum/events
      apiVersion: 0.0.9
      language: wasm/assemblyscript
      entities:
        - RedactShieldedNative
      abis:
        - name: ConfidentialETH
          file: ../abis/redact/ConfidentialETH.json
      eventHandlers:
        - event: ShieldedNative(indexed address,indexed address,uint256)
          handler: handleShieldedNative
          receipt: true
      file: ../src/redact/confidential-eth.ts
```

subgraph/src/redact/confidential-eth.ts
```diff
@@ -0,0 +1,103 @@
...
+  shield.to = event.params.to;
+  shield.value = event.params.value;
+
+  shield.blockNumber = event.block.number;
+  shield.timestamp = event.block.timestamp;
+  shield.txHash = event.transaction.hash;
+  if (event.receipt !== null) {
+    shield.gasUsed = event.receipt!.gasUsed;
+  }
+  shield.gasPrice = event.transaction.gasPrice;
```
0xmad, line 106: New lines between if

subgraph/src/redact/confidential-eth.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  return stats;
+}
+
+function createOrLoadShieldedNativeStats(operationStatsId: string): RedactShieldedNativeStats {
+  let stats = RedactShieldedNativeStats.load(operationStatsId);
+
+  if (stats == null) {
+    stats = new RedactShieldedNativeStats(operationStatsId);
+    stats.tokenAddress = Bytes.fromHexString("0x0000000000000000000000000000000000000000");
```
0xmad: Why do you use zero address?

subgraph/package.json
```diff
@@ -2,9 +2,13 @@
...
-    "codegen": "graph codegen",
-    "build": "graph build",
-    "deploy": "graph deploy --node https://api.studio.thegraph.com/deploy/ private-transfers-benchmarks",
+    "pre-codegen": "tsx scripts/merge-schema.ts",
+    "codegen:mainnet": "pnpm run pre-codegen && graph codegen ./configuration/subgraph.mainnet.yaml",
+    "codegen:sepolia": "pnpm run pre-codegen && graph codegen ./configuration/subgraph.sepolia.yaml",
+    "build:mainnet": "pnpm run pre-codegen && graph build ./configuration/subgraph.mainnet.yaml",
+    "build:sepolia": "pnpm run pre-codegen && graph build ./configuration/subgraph.sepolia.yaml",
+    "deploy:mainnet": "graph deploy  private-transfers-mainnet-benchmarks ./configuration/subgraph.mainnet.yaml",
+    "deploy:sepolia": "graph deploy  private-transfers-sepolia-benchmarks ./configuration/subgraph.sepolia.yaml",
```
0xmad: Remove additional spaces

subgraph/scripts/merge-schema.ts
```diff
@@ -0,0 +1,16 @@
+import { readdirSync, readFileSync, writeFileSync } from "fs";
```
0xmad: Let’s use async methods

subgraph/package.json
```diff
@@ -2,9 +2,13 @@
   "name": "private-transfers-benchmarks",
   "license": "UNLICENSED",
   "scripts": {
-    "codegen": "graph codegen",
-    "build": "graph build",
-    "deploy": "graph deploy --node https://api.studio.thegraph.com/deploy/ private-transfers-benchmarks",
+    "pre-codegen": "tsx scripts/merge-schema.ts",
```
0xmad: Let’s use hooks instead and have one codegen command and specify params for network specific

0xmad (commented):
@NicoSerranoP thanks, left some comments

subgraph/configuration/subgraph.mainnet.yaml
NicoSerranoP: Yes! I put a mention in the PR description:
> Create 2 different subgraph.yml files for the different networks (mainnet for Railgun and sepolia for Redact). We cannot have [more than one network per subgraph.yml file ](https://thegraph.com/docs/en/subgraphs/developing/creating/subgraph-manifest/#subgraph-capabilities). I am not sure if we can [create a subgraph.yml template file like in maci-platform](https://github.com/privacy-ethereum/maci-platform/blob/ca078b796740b82dcadc18246bc637ec706e6fcf/packages/subgraph/templates/subgraph.template.yaml) due to the difference between number of events. We could leave it as a task for another PR.


When putting two dataSources in the same subgraph.yaml as you suggest and run `NETWORK=mainnet pnpm run codegen` we get the following error:
```bash
✔ Apply migrations
✖ Failed to load subgraph from configuration/subgraph.mainnet.yaml: Error in configuration/subgraph.mainnet.yaml:

  Path: /
  Conflicting networks used in data sources and templates:
    Data sources and templates using 'mainnet':
      - RailgunSmartWallet
    Data sources and templates using 'sepolia':
      - ConfidentialETH
  Recommendation: Make all data sources and templates use the same network name.
```

subgraph/src/redact/confidential-eth.ts
NicoSerranoP: I used the zero address to differentiate native ETH shield from other ERC20 shields. By looking at your question I realized that I was using `RedactShieldedNativeStats` to save the stats so I rename it to `RedactShieldedStats`. 

That way we keep the same structure as Railgun and we differentiate ERC20 from native ETH by using the zero address

0xmad: If there is no address for eth, there is no need to add it to entity

subgraph/scripts/merge-schema.ts
JohnGuilding, line 1: So we cannot have multiple schemas and they need to be combined into one?

Agree with splitting files per project btw

subgraph/src/redact/confidential-eth.ts
```diff
@@ -0,0 +1,109 @@
...
+    stats.totalValue = BigInt.fromI32(0);
+    stats.operationStats = operationStatsId;
+
+    stats.save();
+  }
+
+  return stats;
+}
+
+export function handleShieldedNative(event: ShieldedNativeEvent): void {
```
JohnGuilding, line 54: Beginner question, we know that this is a shield event because graphql picks up the emitted shield event?

In other words, my question is, what is the mechanism through which graphql finds the right transaction?

subgraph/src/redact/confidential-eth.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  stats.save();
+
+  shield.from = event.params.from;
+  shield.to = event.params.to;
+  shield.value = event.params.value;
+
+  shield.blockNumber = event.block.number;
+  shield.timestamp = event.block.timestamp;
+  shield.txHash = event.transaction.hash;
```
JohnGuilding, line 100: Why are we gathering all of these details? Why are they necessary?

subgraph/src/redact/confidential-eth.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  if (event.receipt !== null) {
+    shield.gasUsed = event.receipt!.gasUsed;
+  }
+
+  shield.gasPrice = event.transaction.gasPrice;
+
+  shield.save();
+
+  if (shieldStats !== null) {
```
JohnGuilding: Why isn't this done in same block as above?

JohnGuilding (commented):
Left a few questions. Still learning about subgraph use so think it's best to wait on Anton's approval

subgraph/src/redact/confidential-eth.ts
```diff
@@ -0,0 +1,109 @@
+import { BigInt, Bytes } from "@graphprotocol/graph-ts";
+
+import { ShieldedNative as ShieldedNativeEvent } from "../../generated/ConfidentialETH/ConfidentialETH";
```
JohnGuilding, line 3: I'm on the fence about this. I'm wondering ShieldedNative is confusing naming, wdyt about `ShieldedEth` to be explicit?

subgraph/scripts/merge-schema.ts
NicoSerranoP: Yes, `graph` CLI requires a single schema.graphql file so we needed a script to merge different files into one

subgraph/src/redact/confidential-eth.ts
NicoSerranoP: Good question! My thought process was the following:

1. Look at the subgraph.yaml file which has all the configurations. We need one config file per network so in our case we have the directory `/configuration` with `subgraph.mainnet.yaml` and `subgraph.sepolia.yaml`. In the future we could add `.scroll.yaml` or `subgraph.arbitrum.yaml`. 

2. The config file defines a smart contract address (in our case the protocol address), a start block (usually when the contract was deployed) and a list of eventHandlers. Each event in this list has an event definition (e.g.  `Shield(uint256,uint256,(bytes32,(uint8,address,uint256),uint120)[], ...)`) and a handler. There is also a `receipt: true` which makes the subgraph return the receipt info + tx info (only tx info is returned by default)

3. The config file is telling the subgraph: "starting from this block, check this address and if any of the events in the list has been emitted, send that data to the event handler function to process it".

4. So in Redact (deployed in Sepolia) we define an event `ShieldedNative(indexed address,indexed address,uint256)` and send it to `handleShieldedNative`

5. In the `subgraph/src/redact/confidential-eth.ts` file, the main function is `handleShieldedNative`. It loads different stats and saves the info with the new processed event.

6. The stats accumulate totalCount, totalGasUsed, totalValue and they are separated into different entities:
    6.1. `RedactProtocolStats`: the whole protocol data (e.g. "How many txs are in Redact, total gas used in Redact, total value sent in Redact")
    6.2. `RedactOperationStats`: the operation data (e.g. "How many shield txs are in Redact, total gas used in shield txs in Redact, total value sent in shield txs in Redact")
    6.3. `RedactShieldedStats`: the operation-asset level data (e.g. "How many shield native ETH txs are in Redact, total gas used in shield native ETH txs in Redact, total value sent in shield native ETH txs in Redact")
    6.4. `RedactShieldedNative`: the atomic event data (e.g. saves all info of a shield native ETH emitted event)
    
    
 TLDR:
 
 1. check the `subgraph.yaml` file
 2. check the `handlerEVENT` function in `/src/PROTOCOL/`
 3. check the `schema.graphql` file to cross-check with the types in the handler function

subgraph/src/redact/confidential-eth.ts
NicoSerranoP: This data is the event data + tx data + receipt data. 

### I am sure
- `shield.txHash`: will help to identify different events emitted in the same transaction. This way we would be able to map Hinkal or the Redact internal transfers
- `shield.value`: we need it to know how much is being shielded. It will depend a lot in each protocol and their implementation. In this case we could present a metric like "Redact has shielded N native ETH"

### I am not 100% sure
-  `shield.from`
- `shield.to`
- `shield.blockNumber`
- `shield.timestamp`

It could be useful in the future (e.g. give me metrics from in this blockNumber range). We are also saving it on Railgun. If we hit a storage limit we could reduce it but I think we should not touch it now

subgraph/src/redact/confidential-eth.ts
NicoSerranoP: You are right, I just reorganized the blocks in a more readable order:

1. protocol stats
2. operation stats
3. operation-assets stats
4. event info

subgraph/src/redact/confidential-eth.ts
NicoSerranoP: We are writing the schemas based on the event names:

- Railgun emits `Shield` event so the schema is `RailgunShield`
- Redact emits `ShieldedNative` event so the schema is `RedactShieldedNative` 

*We need to put the protocol name at the beginning so we do not confuse events with the same name but different. The final `schema.graphql` file will be a single long file with all defined schemas

I was initially using the name `EncryptETH` because Redact is a FHE confidential token but I realized that I will mess up with this naming convention that will be useful when we get multiple projects in the repo

subgraph/src/redact/confidential-eth.ts
NicoSerranoP: I just removed the zero address and the native ETH will not have a token address. The id for the `RedactShieldedStats` is "native-eth". Worth pointing out that we are capturing the following:

1. protocol stats (e.g. total metrics for Redact protocol) with `RedactProtocolStats`
2. operation stats (e.g. metrics for Redact shield operations) with `RedactOperationStats`
3. operation-asset stats (e.g. metrics for Redact shield native ETH operations) with `RedactShieldedStats`
4. event info (e.g. the specific Redact ShieldedNative event emitted) with `RedactShieldedNative`

subgraph/src/redact/confidential-eth.ts
0xmad: ShieldedNative is a more generic one. If it's deployed on different networks (like BNB) it won't be ETH anymore.

subgraph/src/redact/confidential-eth.ts
```diff
@@ -0,0 +1,109 @@
...
+
+    if (event.receipt !== null) {
+      shieldStats.totalGasUsed = shieldStats.totalGasUsed.plus(event.receipt!.gasUsed);
+    }
+
+    shieldStats.save();
+  }
+
+  if (shieldStats !== null) {
+    const shieldedStats = createOrLoadShieldedStats(null, shieldStats.id);
```
0xmad: Can we rename `shieldedStats` or `shieldStats`? They look similar.

subgraph/src/redact/confidential-eth.ts
JohnGuilding: Understood and current naming makes sense

subgraph/src/redact/confidential-eth.ts
JohnGuilding: Ok I've [added an issue](https://github.com/privacy-ethereum/private-transfers-benchmarks/issues/177) to revisit this. I don't see the value on gathering data for no reason. Plus we didn't do that with original benchmarks logic. 

> shield.value: we need it to know how much is being shielded. It will depend a lot in each protocol and their implementation. In this case we could present a metric like "Redact has shielded N native ETH"

We should define what we want to gather. At the moment "Redact has shielded N native ETH" is not under scope for benchmarks. We can definitely expand, but should increase scope with intention.

JohnGuilding: @0xmad why did we start gathering this data with Railgun? Some of it seems unnecessary for the current scope of cost of shields, private transfer, unshields, and anonymity set sizes

subgraph/src/redact/confidential-eth.ts
JohnGuilding: Very helpful explanation thank you :)

subgraph/src/redact/confidential-eth.ts
JohnGuilding: Btw I think we should revisit the question of scope of benchmarks given extra dev resources, scalability benefits from subgraph, ease of gathering new data. Wonder what could be interesting to consider?

(But also just want to make sure we're being intentional with any changes)

0xmad: All of these values are required in case if you want to calculate metric for custom periods (1 week, 7 weeks and etc) and if you want to filter/query by specific value/field (token address).

0xmad (approved):
@NicoSerranoP thanks!

PR #176, opened 2026-04-22
feat(benchmarks): get base fee per gas + prevent benchmark CI errors

### LINKED ISSUE #175: Fix CI failures
**Description:**
CI is failing consistently with:
`Details: Your app has exceeded its compute units per second capacity. If you have retries enabled, you can safely ignore this message.`

**Some solutions:**
* We could find a way to increase limits or optimise the failing calls so they don't fail
* We could stop failing benchmarks and leave others running

**Acceptance Criteria:**
- CI stops failing on every run

NicoSerranoP:
1. Get the base fee per gas in a specific block window (7 days, 30 days, etc). It fetches the block info in an hourly sampled and computes the average. Fix https://github.com/privacy-ethereum/private-transfers-benchmarks/issues/150
2. Disable the benchmarking CI because we are migrating to subgraph. Fix https://github.com/privacy-ethereum/private-transfers-benchmarks/issues/175

### DIFF
--- a/gas-benchmarks/src/utils/rpc.ts
+++ b/gas-benchmarks/src/utils/rpc.ts
@@ -3,17 +3,19 @@ import {
   encodeEventTopics,
   http,
   toHex,
+  type Block,
   type Chain,
   type Hash,
   type Log,
   type PublicClient,
   type TransactionReceipt,
 } from "viem";
-import { mainnet, scroll } from "viem/chains";
+import { mainnet, scroll, sepolia } from "viem/chains";
 
 import type {
   EthGetBlockReceiptsSchema,
   GetAllLogsInput,
+  GetBaseFeePerGasAverageInDaysWindowInput,
   GetBlockWindowInput,
   GetBlockWindowOutput,
   GetValidEthTransfersInput,
@@ -23,18 +25,22 @@ import type {
 
 import {
   ETH_RPC_URL,
+  SEPOLIA_RPC_URL,
   SCROLL_RPC_URL,
   BLOCK_RANGE,
   MAX_SAMPLES,
   BLOCK_WINDOW_ETHEREUM_1_WEEK,
   BLOCK_WINDOW_SCROLL_1_WEEK,
+  BATCH_SIZE_FOR_RPC_CALLS,
+  DELAY_BETWEEN_BATCHES,
 } from "./constants.js";
-import { isNativeTransfer } from "./utils.js";
+import { isNativeTransfer, sleep } from "./utils.js";
 
 /** Pre-configured RPC clients keyed by chain ID */
 const clients: Record<number, PublicClient | undefined> = {
   [mainnet.id]: createPublicClient({ chain: mainnet, transport: http(ETH_RPC_URL, { batch: true }) }),
   [scroll.id]: createPublicClient({ chain: scroll, transport: http(SCROLL_RPC_URL, { batch: true }) }),
+  [sepolia.id]: createPublicClient({ chain: sepolia, transport: http(SEPOLIA_RPC_URL, { batch: true }) }),
 };
 
 /** Returns the RPC client for a given chain */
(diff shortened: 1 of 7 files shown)


gas-benchmarks/src/utils/rpc.ts
JohnGuilding, line 1: Code simplification idea:

```ts
 const HOURS = 24 * 7;                                                                                                                              
                                                                                                                                                     
  const blocksPerHour = (chainId: number): number => {                                                                                               
    switch (chainId) {                                                                                                                               
      case mainnet.id:                                                                                                                               
      case sepolia.id:                                                                             
        return 300; // 3600s / 12s                                                                                                                   
      case scroll.id:
        return 3600; // 3600s / 1s                                                                                                                   
      default:                                                                                     
        throw new Error(`No block time configured for chain ID: ${chainId}`);
    }                                                                                                                                                
  };
                                                                                                                                                     
  export const getAverageBaseFeePerGas = async (chain: Chain, latestBlock: bigint): Promise<bigint> => {
    const client = getClient(chain);
    const step = blocksPerHour(chain.id);                                                                                                            
   
    const blockNumbers = Array.from(                                                                                                                 
      { length: HOURS },                                                                           
      (_, index) => latestBlock - BigInt(index * step),
    );                                                                                                                                               
   
    const blocks = await Promise.all(                                                                                                                
      blockNumbers.map((blockNumber) => client.getBlock({ blockNumber })),                         
    );                                                                                                                                               
   
    const baseFees = blocks                                                                                                                          
      .map((block) => block.baseFeePerGas)                                                         
      .filter((baseFeePerGas): baseFeePerGas is bigint => baseFeePerGas !== null);
                                                                                                                                                     
    if (baseFees.length === 0) throw new Error(`No base fees for chain ID: ${chain.id}`);
                                                                                                                                                     
    return baseFees.reduce((sum, baseFeePerGas) => sum + baseFeePerGas, 0n) / BigInt(baseFees.length);                                               
  };     
```

If we need to pass different windows in:

```ts 
const HOURS_PER_DAY = 24;

// ...

  export const getAverageBaseFeePerGas = async (                                                                                                     
    chain: Chain,
    latestBlock: bigint,                                                                                                                             
    windowDays: number,                                                                            
  ): Promise<bigint> => {                                                                                                                            
    const client = getClient(chain);
    const step = blocksPerHour(chain.id);                                                                                                            
    const samples = windowDays * HOURS_PER_DAY;                                                    
                                                                                                                                                     
    const blockNumbers = Array.from(                                                                                                                 
      { length: samples },                                                                                                                           
      (_, index) => latestBlock - BigInt(index * step),                                                                                              
    );       
// ...
}
```

JohnGuilding (approved):
Niceeeee.

Added a code simplification idea

gas-benchmarks/src/utils/rpc.ts
```diff
@@ -193,3 +194,53 @@ export const getValidEthTransfers = async ({
...
+
+  const blocksInAnHour = (60 * 60) / blockTimeInSeconds;
+  const step = BigInt(Math.max(1, Math.floor(blocksInAnHour)));
+  const totalHourlySteps = Number((scanStart - scanEnd) / step);
+
+  const blocks = Array.from({ length: totalHourlySteps + 1 }, (_, index) => scanStart - BigInt(index) * step).filter(
+    (blockNumber) => blockNumber >= scanEnd,
+  );
+
+  return blocks;
```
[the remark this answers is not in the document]
NicoSerranoP: With the simplification idea we start at the last block and we go back by steps until the sample is completed. There is no startBlock. The endBlock is the lastBlock which is the first number in the array

gas-benchmarks/src/utils/rpc.ts
NicoSerranoP: I like this better. I was fixed on maintaining the blockWindow concept but the windowDays is a lot better and simpler to read. I just implemented it

PR #186, opened 2026-05-01
feat(subgraph): added fluidkey

NicoSerranoP:
# What this PR does

1. Migrates Fluidkey to subgraph
2. Fluidkey deposit/publicToStealth is a normal token transfer to an address (native ETH or ERC20). That is why I decided to get an approximate cost of the gas used for these types of transactions. I left a chat I had with Claude in order to find the exact gas cost. According to Claude it is not possible and there is no official documentation that indicates that number. We are using approximate in the `gas-benchmarks` package
3. Example of how to read subgraph data in order to write it in the `benchmarks.json` file. In the future I will modify the script to save this data directly in the evaluation json files.

### Notice 🔴 
Please note that https://github.com/privacy-ethereum/private-transfers-benchmarks/pull/183 needs to be merged first

## Confirmation

> [!IMPORTANT]


subgraph/src/redact/confidential-eth.ts
```diff
@@ -107,3 +144,41 @@ export function handleShieldedNative(event: ShieldedNativeEvent): void {
...
 }
+
+export function handleUnshielded(event: UnshieldedEvent): void {
+  const stats = createOrLoadProtocolStats();
+  saveProtocolStats(event);
+
+  const operationId = stats.unshielded;
+  saveOperationStats(operationId, event);
+
+  const unshieldedStats = createOrLoadUnshieldedStats(operationId);
```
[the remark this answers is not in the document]
NicoSerranoP: These are changes of https://github.com/privacy-ethereum/private-transfers-benchmarks/pull/183

subgraph/tests/redact/redact-confidential-eth-utils.ts
```diff
@@ -13,3 +13,13 @@ export function createShieldedNativeEvent(from: Address, to: Address, value: Big
 
   return event;
 }
+
+export function createUnshieldedEvent(to: Address, amount: Bytes): Unshielded {
+  const event = changetype<Unshielded>(newMockEvent());
+  event.parameters = [];
+
+  event.parameters.push(new ethereum.EventParam("to", ethereum.Value.fromAddress(to)));
+  event.parameters.push(new ethereum.EventParam("amount", ethereum.Value.fromBytes(amount)));
```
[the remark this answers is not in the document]
NicoSerranoP: Fixed in https://github.com/privacy-ethereum/private-transfers-benchmarks/pull/183

gas-benchmarks/src/utils/constants.ts
```diff
@@ -85,3 +85,39 @@ export const BLOCK_WINDOW_SCROLL_1_WEEK = 604_800n;
...
+ * Shared among EF's Claude account
+ */
+
+/**
+ * Gas cost of a WETH ERC20 token transfer in EVM
+ */
+export const WETH_ERC20_TRANSFER = {
+  new: 46_900,
+  existing: 29_100,
+};
```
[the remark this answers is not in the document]
NicoSerranoP: We need to discuss about the approximate ERC20 values in this PR or later. I will leave the link to the Claude chat in case the team wants to check it out

gas-benchmarks/src/utils/types.ts
```diff
@@ -39,6 +39,6 @@ export interface ProtocolConfig {
 /** Fee metrics */
 export interface FeeMetrics {
   averageGasUsed: bigint | number | "no-data";
-  averageGasPrice: bigint | number | "no-data";
-  averageTxFee: bigint | number | "no-data";
+  averageGasPrice?: bigint | number | "no-data";
+  averageTxFee?: bigint | number | "no-data";
```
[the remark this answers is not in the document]
NicoSerranoP: We might need to only focus on averageGasUsed for scope. We can include the other metrics later

gas-benchmarks/package.json
```diff
@@ -7,6 +7,7 @@
   "scripts": {
     "prebenchmark": "pnpm run codegen",
     "benchmark": "tsx src/index.ts",
+    "fluidkey": "tsx src/fluidkey/new.ts",
```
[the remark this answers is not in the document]
NicoSerranoP: This is a leftover. I am deleting it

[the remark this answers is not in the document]
NicoSerranoP: This is a leftover. I am deleting it

subgraph/src/fluidkey/smart-account-relayer.ts
```diff
@@ -0,0 +1,97 @@
...
+  return stats;
+}
+
+export function handleOperationRelayed(event: OperationRelayedEvent): void {
+  const stats = createOrLoadProtocolStats();
+
+  stats.totalTxCount = stats.totalTxCount.plus(BigInt.fromI32(1));
+
+  if (event.receipt !== null) {
+    stats.totalGasUsed = stats.totalGasUsed.plus(event.receipt!.gasUsed);
```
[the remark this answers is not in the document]
NicoSerranoP: You are right, I just added a if block

gas-benchmarks/src/index.ts
```diff
@@ -83,12 +79,7 @@ await db.update((data) => {
   };
 
   // eslint-disable-next-line no-param-reassign
-  data[`${fluidkey.name}_${fluidkey.version}`] = {
-    shield_eth: fluidkeyMetrics.shieldEth,
-    shield_erc20: fluidkeyMetrics.shieldErc20,
-    transfer_eth: fluidkeyMetrics.transferEth,
-    transfer_erc20: fluidkeyMetrics.transferErc20,
-  };
+  data[fluidkey.id] = fluidkeyMetrics;
```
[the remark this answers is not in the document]
NicoSerranoP: You are right but we are changing how we are gathering the metrics and how we will present them. I am expecting to have other protocols change/migrate to follow this new convention.

gas-benchmarks/src/subgraph/index.ts
```diff
@@ -1,15 +1,105 @@
...
+      }
+    }
+  `;
+
+  async getResults() {
+    if (this.results === null) {
+      const client = new GraphQLClient(SUBGRAPH_URL);
+
+      this.results = await client.request<TRootQuery>(this.rootQuery);
+    }
```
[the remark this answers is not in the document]
NicoSerranoP: Fixed

gas-benchmarks/src/fluidkey/index.ts
```diff
@@ -1,90 +1,53 @@
...
+    const { fluidkeyProtocolStats } = results;
+
+    const stealthToPublic = fluidkeyProtocolStats?.stealthToPublic;
+    const totalGasUsed = stealthToPublic ? Number(stealthToPublic.totalGasUsed) : null;
+    const totalCount = stealthToPublic ? Number(stealthToPublic.totalCount) : null;
+
+    const isThereData = totalGasUsed !== null && totalCount !== null && totalCount > 0;
+
+    const gasUsedAverage = isThereData ? totalGasUsed / totalCount : "no-data";
+
```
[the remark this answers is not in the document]
NicoSerranoP: Fixed

gas-benchmarks/src/index.ts
```diff
@@ -19,10 +19,6 @@ await db.read();
 
 const start = Date.now();
 
-// const client = new GraphQLClient(SUBGRAPH_URL);
-// const result = await client.request<TRootQuery>(RootQuery);
-// console.log(result);
-
```
0xmad, line 25: Do not remove it

gas-benchmarks/src/subgraph/index.ts
```diff
@@ -1,19 +1,41 @@
...
+      const client = new GraphQLClient(SUBGRAPH_URL);
+
+      this.results = await client.request<TRootQuery>(RootQuery);
+    }
+
+    return this.results;
+  }
+}
+
+export const subgraph = new Subgraph();
```
0xmad: Why do you need it?
If you want to use caching, it should be done differently (not just setting class property)

gas-benchmarks/src/fluidkey/index.ts
```diff
@@ -1,90 +1,53 @@
...
+      },
+      stealthToPublicUSDC: {
+        averageGasUsed: gasUsedAverage,
+      },
+      publicToStealthUSDT: {
+        averageGasUsed: USDT_ERC20_TRANSFER.new,
+      },
+      stealthToPublicUSDT: {
+        averageGasUsed: gasUsedAverage,
+      },
```
0xmad, line 38: Why only 3 hardcoded tokens?

gas-benchmarks/src/subgraph/index.ts
NicoSerranoP: What would be a better approach? My idea was to only request the Subgraph data once and use those results in different files

gas-benchmarks/src/fluidkey/index.ts
NicoSerranoP: I think WETH, USDC and USDT are the most popular ones. We could add more tokens in the future. I think it is super important we review how we are calculating those gas costs

gas-benchmarks/src/subgraph/index.ts
0xmad: When you run benchmarks, it initiates the same request everytime. The better approach would be to use caching module for graphql (similar to apollo client).

gas-benchmarks/src/fluidkey/index.ts
NicoSerranoP: I just added a utils/README.md file explaining the hardcoded values:

- Why these specific tokens were selected
- How we are calculating the gas units used

gas-benchmarks/src/subgraph/index.ts
NicoSerranoP: I see your point. My idea is to avoid request GraphQL data multiple times in a single `pnpm run benchmark` run. At the moment we are only requesting it once in `gas-benchmarks/src/fluidkey/index.ts` but in the future we will have +15 protocols and that will mean +15 requests of GraphQL data per `pnpm run benchmark` run.

I feel that targeting more complex cache solutions (e.g. using Apollo Client library) is an overkill. [`GraphQLClient`](https://github.com/graffle-js/graffle/tree/graphql-request#what-sets-graphql-request-apart-from-other-clients-like-apollo-relay-etc) is "the most minimal and simplest way to use GraphQL client" and "doesn't have a built-in cache".

If we really require a more complex cache solution I could create an issue to tackle it later 😊

0xmad: In this case, let's not add this functionality here because we can hit API limits really fast. This should be added with caching.

NicoSerranoP: I do not get why we will hit the API limit if we are saving the results inside the class. I'll change the `gas-benchmarks/src/fluidkey/index.ts` to

```javascript
const client = new GraphQLClient(SUBGRAPH_URL);
const result = await client.request<TRootQuery>(RootQuery);
```


But with more protocols we will have one request per protocol. By using the class we will use the in-memory saved results

0xmad: Every time you lauch benchmark command, you create a new request. My suggestion is to avoid calling subgraph until caching solution is implemented (see how it's implemented for other migrated benchmarks, there are no any calls to subgraph).

NicoSerranoP: So you want to prevent calling the subgraph at all. I will leave the metrics with "no-data" as a placeholder for now

0xmad (approved):
@NicoSerranoP thanks!

subgraph/src/fluidkey/smart-account-relayer.ts
```diff
@@ -0,0 +1,101 @@
...
+export function handleOperationRelayed(event: OperationRelayedEvent): void {
+  if (!event.params.success) {
+    return;
+  }
+
+  const stats = createOrLoadProtocolStats();
+
+  stats.totalTxCount = stats.totalTxCount.plus(BigInt.fromI32(1));
+
+  if (event.receipt !== null) {
```
JohnGuilding, line 58: Why do we need this check if we have `receipt: true`? `receipt: true` doesn't force a receipt? 

```
      eventHandlers:
        - event: OperationRelayed(indexed uint256,indexed bool)
          handler: handleOperationRelayed
          receipt: true
```

gas-benchmarks/src/utils/types.ts
```diff
@@ -39,6 +39,6 @@ export interface ProtocolConfig {
 /** Fee metrics */
 export interface FeeMetrics {
   averageGasUsed: bigint | number | "no-data";
-  averageGasPrice: bigint | number | "no-data";
-  averageTxFee: bigint | number | "no-data";
+  averageGasPrice?: bigint | number | "no-data";
```
JohnGuilding, line 42: Why are we marking these as optional?

gas-benchmarks/src/utils/README.md
JohnGuilding, line 1: I'm currently thinking this approach feels a little too arbitrary and error prone.

I haven't reviewed opcodes in detail but some review from Claude:
  - "EIP-2029" doesn't exist — SLOAD/access costs are EIP-2929, SSTORE semantics EIP-2200.
  - Building-blocks row says "LOG2" — Transfer is LOG3 (1,756), as the per-token tables already use.
  - USDT section quotes transferFrom; users call transfer (extra fee SLOADs) — existing figure too low.
  - USDC section cites FiatTokenV1; live proxy is V2.2 (fewer SLOADs) — wrong source.
  - Constants don't match the README's totals (50,900 vs 50,864, etc.), rounded inconsistently.
  - existing column is inapplicable (stealth deposits always hit a zero slot) and unused in code.
  - "New recipient" labelled as new address; the swing is actually a zero balance slot.
  - Sender-empties-balance case omitted (nonzero→zero SSTORE + ~4,800 refund).
  - "Misc opcodes ~400–600" is an unverified plug; ignores dispatcher, modifiers, SafeMath, memory expansion.
  - L2 tokens listed but model is mainnet-execution-only — no L1 data cost; averageTxFee will mislead.
  - Calldata booked as flat 608 gas though it's amount-dependent.
  - Provenance is a claude.ai/share/... link — not pinned, not reproducible.
  - No real tx hash / gasUsed anchor, no drift test — should be measured (fork test or averaged receipts), not hand-tallied.

Some alternative and simple approaches also outlined by Claude:

### 1. eth_estimateGas against the token contract — one viem call per token.
Not an "average", but a real, current number, and it naturally exercises the 20k-SSTORE fresh-recipient case if you point to at a brand-new address. Also auto-fixes the "we cited FiatTokenV1 but V2.2 is deployed" problem.
```ts
const data = encodeFunctionData({ abi: erc20Abi, functionName: "transfer", args: [freshRandomAddr, 1n] });
const gas = await client.estimateGas({ account: knownHolder, to: TOKEN, data }); // or use stateOverride to fund a dummy holder. 
```
Four lines, four tokens, done. ETH transfer is just 21_000n — keep that hardcoded.

### 2. Reuse what the repo already does — getValidTransactions over a short block window + getAverageMetrics.
This gives a genuine real-world average and matches every other protocol's pattern. Add one tiny filter analogous to the existing isNativeTransfer:
```ts
const isSimpleErc20Transfer = (r: TransactionReceipt, token: Address) =>
    r.to?.toLowerCase() === token.toLowerCase() && r.logs.length === 1;
```
Filter Transfer logs from the token contract → fetch receipts → keep the one-log ones → getAverageMetrics. Caveat: this blends fresh + existing recipients by however often each occurs on mainnet, so it'll read a bit lower than the stealth-deposit case — document that, or don't worry about it.

### 3. Foundry fork test with snapshot gas
If forge is already wired up here. Most precise, exact current bytecode:
```solidity
vm.createSelectFork(MAINNET_RPC);
vm.prank(whale);
vm.startSnapshotGas("usdc");
IERC20(USDC).transfer(makeAddr("fresh"), 1e6);
vm.stopSnapshotGas();
```
~5 lines per token. Adds a test harness if the package is currently pure-TS, so only worth it if Foundry's already there.

### 4. Zero-code fallback: hardcode from real receipts, not hand-tallied opcodes.
cast receipt <hash> on ~5 real transfers per token, average them, put the tx hashes in the JSDoc. Still stale-prone, but at least every number is something that actually happened on-chain. Strictly better than the current README method; strictly worse than 1–3 because nothing refreshes it.

JohnGuilding: Wdyt?

I'm thinking option 1 or 2 sound good.

To save time I'm wondering if we keep it simple and go with option 1 (or even 3 or 4). Then make it more complex after the deadline. Open to other suggestions though

My thinking is it is better to be accurate and simple, then push inaccurate benchmarks live.

JohnGuilding: I think option 2 may be wrong as it is based on old methodology rather than subgraph

PR #188, opened 2026-05-05
feat(subgraph): add worm

NicoSerranoP:
# What this PR does

1. Added WORM protocol to subgraph. It is worth pointing out that WORM does not have a singular unique event to identify the withdraw transaction (e.g. Withdraw, Unshield, Decrypt). It has a sequence of events emitted by the BETH contract that can be used to identify withdraw transactions

2. created the `logsMatchEvents` function to reuse in future protocol additions

## Confirmation

> [!IMPORTANT]

### DIFF
--- a/subgraph/configuration/subgraph.mainnet.yaml
+++ b/subgraph/configuration/subgraph.mainnet.yaml
@@ -138,6 +138,30 @@ dataSources:
           handler: handlePoolRemoved
       file: ../src/privacy-pools/privacy-pools.ts
 
+  - kind: ethereum
+    name: WETH
+    network: mainnet
+    source:
+      address: "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2"
+      abi: WETH
+      startBlock: 24436025
+    mapping:
+      kind: ethereum/events
+      apiVersion: 0.0.9
+      language: wasm/assemblyscript
+      entities:
+        - WormProtocolStats
+        - WormOperationStats
+        - WormWithdraw
+      abis:
+        - name: WETH
+          file: ../abis/worm/WETH.json
+      eventHandlers:
+        - event: Withdrawal(indexed address,uint256)
+          handler: handleWithdrawal
+          receipt: true
+      file: ../src/worm/weth.ts
+
 templates:
   - kind: ethereum
     name: TornadoInstance
--- a/subgraph/schemas/worm.graphql
+++ b/subgraph/schemas/worm.graphql
@@ -1,0 +1,24 @@
+type WormProtocolStats @entity(immutable: false) {
+  id: ID!
+  totalTxCount: BigInt!
+  totalGasUsed: BigInt!
+
+  withdraw: WormOperationStats!
+}
+
+type WormOperationStats @entity(immutable: false) {
+  id: ID!
+  totalCount: BigInt!
+  totalGasUsed: BigInt!
+}
+
+type WormWithdraw @entity(immutable: true) {
+  id: ID!
+  value: BigInt!
+
+  blockNumber: BigInt!
+  timestamp: BigInt!
+  txHash: Bytes!
+  gasUsed: BigInt!
+  gasPrice: BigInt!
+}
--- a/subgraph/src/utils.ts
+++ b/subgraph/src/utils.ts
@@ -1,0 +1,19 @@
+import { crypto as graphCrypto, ethereum, Bytes } from "@graphprotocol/graph-ts";
+
+export function logsMatchEvents(logs: ethereum.Log[], events: string[]): boolean {
+  if (logs.length !== events.length) {
+    return false;
+  }
+
+  for (let i = 0; i < logs.length; i += 1) {
+    const event = Bytes.fromUTF8(events[i]);
+    const signature = graphCrypto.keccak256(event);
+    const topic = logs[i].topics[0];
+
+    if (!signature.equals(topic)) {
+      return false;
+    }
+  }
+
+  return true;
+}
--- a/subgraph/src/worm/beth.ts
+++ b/subgraph/src/worm/beth.ts
@@ -1,0 +1,106 @@
+import { BigInt, Bytes } from "@graphprotocol/graph-ts";
+
+import { Transfer as TransferEvent } from "../../generated/BETH/BETH";
+import { WormOperationStats, WormProtocolStats, WormWithdraw } from "../../generated/schema";
+import { logsMatchEvents } from "../utils";
+
+export const WITHDRAW_EVENTS = [
+  "Transfer(address,address,uint256)",
+  "Transfer(address,address,uint256)",
+  "Transfer(address,address,uint256)",
+  "Transfer(address,address,uint256)",
+  "Transfer(address,address,uint256)",
+  "Approval(address,address,uint256)",
+  "Transfer(address,address,uint256)",
+  "Approval(address,address,uint256)",
+  "Transfer(address,address,uint256)",
+  "Transfer(address,address,uint256)",
+  "Swap(address,address,int256,int256,uint160,uint128,int24,uint24,uint24)",
+  "Approval(address,address,uint256)",
+  "Withdraw(address,uint256)",
+  "Approval(address,address,uint256)",
+  "Transfer(address,address,uint256)",
+  "Transfer(address,address,uint256)",
+  "Approval(address,address,uint256)",
+  "Transfer(address,address,uint256)",
+  "RewardDeposited(address,uint256,uint256)",
+  "Approval(address,address,uint256)",
+];
+
+export const INDEX_FOR_WETH_WITHDRAW_EVENT = 12;
+export const INDEX_FOR_TRANSFER_TO_FINAL_RECIPIENT_EVENT = 14;
+
+function createOrLoadProtocolStats(): WormProtocolStats {
+  const id = "worm-protocol-stats";
+  let stats = WormProtocolStats.load(id);
+
+  if (stats == null) {
+    stats = new WormProtocolStats(id);
+    stats.totalTxCount = BigInt.fromI32(0);
+    stats.totalGasUsed = BigInt.fromI32(0);
+
+    const withdraw = new WormOperationStats(`${id}-withdraw`);
+    stats.withdraw = withdraw.id;
+
+    withdraw.totalCount = BigInt.fromI32(0);
+    withdraw.totalGasUsed = BigInt.fromI32(0);
+
+    withdraw.save();
+    stats.save();
+  }
+
+  return stats;
+}
+
+export function handleTransfer(event: TransferEvent): void {
+  const id = `${event.transaction.hash.toHex()}`;
+
+  let withdraw = WormWithdraw.load(id);
+
+  if (withdraw != null) {
+    return; // tx already processed
+  }
+
+  if (event.receipt === null) {
+    return; // we need the receipt to get the logs/events of the tx
+  }
+
+  const isWithdraw = logsMatchEvents(event.receipt!.logs, WITHDRAW_EVENTS);
+
+  if (!isWithdraw) {
+    return; // not a withdraw tx, ignore
+  }
+
+  // the unwrapped amount of WETH is the amount being sent
+  const amountBytes = event.receipt!.logs[INDEX_FOR_WETH_WITHDRAW_EVENT].topics[2];
+  // 0 tokens ERC20 transfer event indicating receiver. The transfer is a native ETH internal transfer
+  const receiverTopic = event.receipt!.logs[INDEX_FOR_TRANSFER_TO_FINAL_RECIPIENT_EVENT].topics[2];
+  const receiverHex = receiverTopic.toHexString();
+  const receiver = Bytes.fromHexString(`0x${receiverHex.slice(26)}`);
+
+  withdraw = new WormWithdraw(id);
+  withdraw.value = BigInt.fromUnsignedBytes(Bytes.fromUint8Array(amountBytes.reverse()));
+  withdraw.to = receiver;
+
+  withdraw.blockNumber = event.block.number;
+  withdraw.timestamp = event.block.timestamp;
+  withdraw.txHash = event.transaction.hash;
+  withdraw.gasUsed = event.receipt!.gasUsed;
+  withdraw.gasPrice = event.transaction.gasPrice;
+
+  withdraw.save();
+
+  const stats = createOrLoadProtocolStats();
+  stats.totalTxCount = stats.totalTxCount.plus(BigInt.fromI32(1));
+  stats.totalGasUsed = stats.totalGasUsed.plus(event.receipt!.gasUsed);
+
+  const operationStats = WormOperationStats.load(stats.withdraw);
+
+  if (operationStats !== null) {
+    operationStats.totalCount = operationStats.totalCount.plus(BigInt.fromI32(1));
+    operationStats.totalGasUsed = operationStats.totalGasUsed.plus(event.receipt!.gasUsed);
+    operationStats.save();
+  }
+
+  stats.save();
+}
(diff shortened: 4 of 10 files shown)


subgraph/src/redact/confidential-eth.ts
```diff
@@ -107,3 +144,41 @@ export function handleShieldedNative(event: ShieldedNativeEvent): void {
...
+
+  const unshieldedStats = createOrLoadUnshieldedStats(operationId);
+
+  if (event.receipt !== null) {
+    unshieldedStats.totalGasUsed = unshieldedStats.totalGasUsed.plus(event.receipt!.gasUsed);
+  }
+
+  unshieldedStats.operationStats = operationId;
+  unshieldedStats.totalCount = unshieldedStats.totalCount.plus(BigInt.fromI32(1));
+  // TODO: calculate unshieldedStats.totalValue for unshielded events using the Transfer() event
```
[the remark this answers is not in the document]
NicoSerranoP: I will fix this later

gas-benchmarks/src/index.ts
```diff
@@ -83,12 +79,7 @@ await db.update((data) => {
   };
 
   // eslint-disable-next-line no-param-reassign
-  data[`${fluidkey.name}_${fluidkey.version}`] = {
-    shield_eth: fluidkeyMetrics.shieldEth,
-    shield_erc20: fluidkeyMetrics.shieldErc20,
-    transfer_eth: fluidkeyMetrics.transferEth,
-    transfer_erc20: fluidkeyMetrics.transferErc20,
-  };
+  data[fluidkey.id] = fluidkeyMetrics;
```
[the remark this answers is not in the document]
NicoSerranoP: we will change all the format

subgraph/configuration/subgraph.mainnet.yaml
```diff
@@ -57,6 +57,52 @@ dataSources:
...
+      abi: BETH
+      startBlock: 24436025
+    mapping:
+      kind: ethereum/events
+      apiVersion: 0.0.9
+      language: wasm/assemblyscript
+      entities:
+        - WormProtocolStats
+        - WormOperationStats
+        - WormMintCoin
```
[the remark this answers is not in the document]
NicoSerranoP: fixed

gas-benchmarks/src/fluidkey/index.ts
```diff
@@ -1,90 +1,53 @@
...
+        averageGasUsed: NATIVE_ETH_TRANSFER,
+      },
+      stealthToPublicETH: {
+        averageGasUsed: gasUsedAverage,
+      },
+      publicToStealthWETH: {
+        averageGasUsed: WETH_ERC20_TRANSFER.new,
+      },
+      stealthToPublicWETH: {
+        averageGasUsed: gasUsedAverage,
```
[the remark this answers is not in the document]
NicoSerranoP: We will segment it in a future PR

gas-benchmarks/src/subgraph/index.ts
0xmad, line 1: Let's remove changes made to this file

subgraph/src/worm/beth.ts
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  return stats;
+}
+
+export function handleTransfer(event: TransferEvent): void {
+  const id = `${event.transaction.hash.toHex()}`;
+
+  let withdraw = WormWithdraw.load(id);
+  if (withdraw != null) {
```
0xmad, line 60: New line before ifs

subgraph/src/utils.ts
```diff
@@ -0,0 +1,20 @@
...
+  if (logs.length !== events.length) {
+    return false;
+  }
+
+  for (let i = 0; i < logs.length; i += 1) {
+    const event = Bytes.fromUTF8(events[i]);
+    const signature = graphCrypto.keccak256(event).toHex();
+    const topic = logs[i].topics[0].toHex();
+
+    // eslint-disable-next-line eqeqeq
```
0xmad: Why do you need to disable this?

subgraph/tests/worm/utils.ts
```diff
@@ -0,0 +1,131 @@
...
+    // eslint-disable-next-line eqeqeq
+    if (i == INDEX_FOR_WETH_WITHDRAW_EVENT) {
+      // Swap event: topic[1] = contract, topic[2] = amount
+      const contract = BETH_TO_ETH_CONTRACT.toAddress();
+      topics.push(padBytes32(contract));
+      topics.push(bigIntToBytes32(value));
+    }
+
+    // eslint-disable-next-line eqeqeq
+    if (i == INDEX_FOR_TRANSFER_TO_FINAL_RECIPIENT_EVENT) {
```
0xmad: Don't disable eslint

subgraph/configuration/subgraph.mainnet.yaml
```diff
@@ -107,6 +107,30 @@ dataSources:
...
+      language: wasm/assemblyscript
+      entities:
+        - WormProtocolStats
+        - WormOperationStats
+        - WormWithdraw
+      abis:
+        - name: BETH
+          file: ../abis/worm/BETH.json
+      eventHandlers:
+        - event: Transfer(indexed address,indexed address,uint256)
```
0xmad: Why not Withdraw event?

subgraph/schemas/worm.graphql
```diff
@@ -0,0 +1,25 @@
+type WormProtocolStats @entity(immutable: false) {
+  id: ID!
+  totalTxCount: BigInt!
+  totalGasUsed: BigInt!
+
+  withdraw: WormOperationStats!
```
0xmad, line 6: Why only withdraw and no transact?

subgraph/src/worm/beth.ts
```diff
@@ -0,0 +1,104 @@
...
+  "Approval(address,address,uint256)",
+  "Withdraw(address,uint256)",
+  "Approval(address,address,uint256)",
+  "Transfer(address,address,uint256)",
+  "Transfer(address,address,uint256)",
+  "Approval(address,address,uint256)",
+  "Transfer(address,address,uint256)",
+  "RewardDeposited(address,uint256,uint256)",
+  "Approval(address,address,uint256)",
+];
```
0xmad, line 28: You only need `Withdraw` event here. No need to add additional logic with logs parsing here.

0xmad (commented):
@NicoSerranoP thanks, added some comments

subgraph/src/worm/beth.ts
JohnGuilding: How does this sequence of events identify a withdraw transaction out of interest? 

It seems quite random

JohnGuilding: I can see `mintCoin` appears to do a lot of transfers, these are all the events emitted just from `mintCoin`?

gas-benchmarks/src/subgraph/index.ts
NicoSerranoP: Sorry I didn't realize it was there. I just deleted it

subgraph/src/worm/beth.ts
NicoSerranoP: There is a more in-depth explanation in `subgraph/src/worm/README.md` but a TLDR:

In order to mint BETH and withdraw your funds from the WORM protocol you need to interact with the protocol smart contract. This interaction involves minting BETH and transfering to the service provides (relayer, prover) and the receiver.

As you can see in the example tx there is no Withdraw event emitted from the WORM smart contract itself. The Withdraw event in the list is emitted by WETH contract. This tx was executed using their UI and the flow is:

1. mint and send BETH to relayer
2. mint and send BETH to prover
3. mint and send BETH to a DEX
4. swap BETH by WETH
5. withdraw ETH from WETH
6. send ETH to receiver


Steps 4 - 6 are not explicitly written in the contract but they are send as hooks in the function params

subgraph/schemas/worm.graphql
NicoSerranoP: Because WORM only allows you to withdraw from the smart contract. Comparing to other protocols:

1. Deposit -> User burns ETH to a unspendable address (it cost 21k gas units)
2. Withdraw -> Withdraw function benchmarked here
3. Transfer -> There is no internal private transfer logic. You could mint BETH and not swap it for ETH and send BETH as a normal ERC20 token but that is not a protocol internal private transfer

subgraph/configuration/subgraph.mainnet.yaml
NicoSerranoP: I put the answer in the next comment. Let's discuss it below 😊

subgraph/src/utils.ts
NicoSerranoP: This is related to AssemblyScript (language used by TheGraph to run the handlers). AssemblyScript threats `===` an `!==` as direct value and position comparison. So if you have `i == INDEX_FOR_WETH_WITHDRAW_EVENT` you would need `const i = INDEX_FOR_WETH_WITHDRAW_EVENT` directly to be true.

It will be false in these scenarios:

1. `i = 1; i == INDEX_FOR_WETH_WITHDRAW_EVENT)`
2. `i = 2; i == INDEX_FOR_WETH_WITHDRAW_EVENT)`
3. `i = 3; i == INDEX_FOR_WETH_WITHDRAW_EVENT)`
4. ....


So the if inside the for loop will not work if we use `===` or `!==`. It is because AssemblyScript

subgraph/tests/worm/utils.ts
NicoSerranoP: I left a comment above, lets discuss it there 😊

subgraph/src/utils.ts
0xmad: So, don't convert the values to hex string and use `equals` method.

subgraph/schemas/worm.graphql
0xmad: Do we need to measure these ERC20 transactions?

subgraph/src/worm/beth.ts
0xmad: It can be implemented simplier if you listen to WETH events and just filter by BETH address

NicoSerranoP: We could do that but:

1. we already have this solution. We need to publish this asap
2. this solution is generalizable for other protocols that do not emit unique identifiable events

subgraph/schemas/worm.graphql
NicoSerranoP: I don't think so because BETH transfers are not private.

subgraph/src/worm/beth.ts
JohnGuilding: Is that list of events accurate?

Found a few `mintCoin` calls with different number of logs emitted:
1. 20 logs https://etherscan.io/tx/0xf11ea6e0523164e61396abcd88f5d42c8a3c7b0074a89db3df89a7170016d263
2. 20 logs https://etherscan.io/tx/0xfae3dbcc05d7ce2d4cd0e46a67d721b1d218ca08b5f7972fb7325f287239679d
3. 22 logs https://etherscan.io/tx/0x3fa776b375d093228ae1e0a5d29fa17d7730462cbf713a795d43c6d574cf4c88
4. 18 logs https://etherscan.io/tx/0xe876f3bc57a39df8d460b17d19d60d535c3fdaa6272c6e47420232232e48c6b0

0xmad: It's not a generic solution because it can be broken with event reordering. Hinkal solution has similar  issue but there is no way to distinguish operations. For this case, it's easier and more stable to listen to WETH `Withdraw` event and just filter by address. You just need to replace yaml config and and one `if` at the beggining of handler.

NicoSerranoP: It is a tx sent by myself from the UI. This address is a EOA (not smart account). I will dig more into these other txs

subgraph/src/utils.ts
NicoSerranoP: Anton I do not get the suggestion, can you be more explicit? The code looks like:

```javascript

// beth.ts file
export const INDEX_FOR_WETH_WITHDRAW_EVENT = 12;
export const INDEX_FOR_TRANSFER_TO_FINAL_RECIPIENT_EVENT = 14;

// utils.ts file

  for (let i = 0; i < WITHDRAW_EVENTS.length; i += 1) {
    // ...

    // eslint-disable-next-line eqeqeq
    if (i == INDEX_FOR_WETH_WITHDRAW_EVENT) {
      // ...
    }

    // eslint-disable-next-line eqeqeq
    if (i == INDEX_FOR_TRANSFER_TO_FINAL_RECIPIENT_EVENT) {
      // ...
    }
    
    // ...
  }
```

There is no `.equals()` method for integers (i32) in [AssemblyScript](https://www.assemblyscript.org/stdlib/number.html)

0xmad: I left the comment for [subgraph/src/utils.ts](https://github.com/privacy-ethereum/private-transfers-benchmarks/pull/188/files/bddc1fb4d7e97c5037d06687a6acfde8fde6aa08#diff-3dd6898ff040322ced12df56c4bd3d07f9ccfd3b087a2f029826022402cd3bcc), there are no i32 types.

```
const signature = graphCrypto.keccak256(event);
const topic = logs[i].topics[0];

if (!signature.equals(topic)) {
  ...
}
```

0xmad: And there is no problem to use `===` for i32. Try to run this and see it works.

```
const INDEX = 0;

function logsMatchEvents(logs: ethereum.Log[], events: string[]): boolean {
  if (logs.length !== events.length) {
    return false;
  }

  for (let i = 0; i < logs.length; i += 1) {
    if (i === INDEX) {
      return true;
    }
  }

  return false;
}
```

NicoSerranoP: I made the changes (remove eslint disable comments) and added another test (incorrect event in logs) to prove it is working correctly.

subgraph/src/worm/beth.ts
NicoSerranoP: @JohnGuilding those txs have basically the same logic:

1. send fee to prover
2. send fee to relayer
3. send fee to recipient
4. swap to WETH (using hook)
5. sending rewards to the pool


The difference between txs is in the `Transfer` or `Approval` events. I am not sure why this happens but the initial 20 logs tx in the README.md file was sent by myself using WORM UI. My idea was to have the most simple withdraw tx so we can benchmark those.

@0xmad the approach of listening to WETH solves the problem above. The issue is that we will be considering all txs that the WORM contract "Withdraw" from WETH. There could be cases were this txs will be more complex and could gas spike our benchmarks.


I can change it to the listening to WETH approach but I feel it is important to consider this.

JohnGuilding: I think both options work. I'm leaning towards filtering withdraw events from specific contract since it's seems slightly more stable and is simple.

> There could be cases were this txs will be more complex and could gas spike our benchmarks.

This is an interesting point. My understanding of these benchmarks now is that with subgraph, we are capturing all operations where possible, and that could mean operations where the tx is more complex and txs where it is less complex — is this right? 

And then Hinkal transfer benchmarks appear to use old method of choosing simplest of operations (e.g. 1 Commitment + 1 Nullifier = simple internal transfer). Open question for me would be if that is true of Hinkal, perhaps the chosen event filters get all relevant txs and not just the simple ones.

In which case maybe some benchmarked operations are simplest possible and some are not already? 

I don't think it is end of the world if we allow more complex transactions in benchmarks here. Think this is the right thing to be thinking about though regarding benchmark accuracy, and maybe we could signify this difference in a comment or in the UI (readme that gets added to UI?)

0xmad: @NicoSerranoP how does you approach prevent gas spikes?

NicoSerranoP: I just implemented the WETH listener. Can you help me review?

0xmad (approved):
@NicoSerranoP thanks!

PR #190, opened 2026-05-05
feat(subgraph): add hinkal event mappers

0xmad:
# What this PR does

- [x] Add shield/unshield event mappers for erc20 and native tokens
- [x] Remove unused data from entities
- [x] Add correct entity names for subgraph config

## Confirmation

> [!IMPORTANT]

### DIFF
--- a/subgraph/schemas/railgun.graphql
+++ b/subgraph/schemas/railgun.graphql
@@ -43,38 +43,11 @@ type RailgunShield @entity(immutable: true) {
   startPosition: BigInt!
   fees: [BigInt!]!
 
-  commitments: [RailgunCommitment!]! @derivedFrom(field: "shield")
-  ciphertexts: [RailgunShieldCiphertext!]! @derivedFrom(field: "shield")
-
   blockNumber: BigInt!
   timestamp: BigInt!
   txHash: Bytes!
   gasUsed: BigInt!
   gasPrice: BigInt!
-}
-
-type RailgunCommitment @entity(immutable: true) {
-  id: ID!
-  shield: RailgunShield!
-
-  npk: Bytes!
-
-  tokenType: Int!
-  tokenAddress: Bytes!
-  tokenSubID: BigInt!
-
-  value: BigInt!
-}
-
-type RailgunShieldCiphertext @entity(immutable: true) {
-  id: ID!
-  shield: RailgunShield!
-
-  encryptedBundle0: Bytes!
-  encryptedBundle1: Bytes!
-  encryptedBundle2: Bytes!
-
-  shieldKey: Bytes!
 }
 
 type RailgunUnshield @entity(immutable: true) {
@@ -101,27 +74,9 @@ type RailgunTransact @entity(immutable: true) {
   startPosition: BigInt!
   hash: [Bytes!]!
 
-  ciphertexts: [RailgunTransactCiphertext!]! @derivedFrom(field: "transact")
-
   blockNumber: BigInt!
   timestamp: BigInt!
   txHash: Bytes!
   gasUsed: BigInt!
   gasPrice: BigInt!
-}
-
-type RailgunTransactCiphertext @entity(immutable: true) {
-  id: ID!
-  transact: RailgunTransact!
-
-  ciphertext0: Bytes!
-  ciphertext1: Bytes!
-  ciphertext2: Bytes!
-  ciphertext3: Bytes!
-
-  blindedSenderViewingKey: Bytes!
-  blindedReceiverViewingKey: Bytes!
-
-  annotationData: Bytes!
-  memo: Bytes!
 }
--- a/subgraph/src/hinkal/hinkal-pool.ts
+++ b/subgraph/src/hinkal/hinkal-pool.ts
@@ -1,0 +1,357 @@
+import { BigInt, Bytes } from "@graphprotocol/graph-ts";
+
+import { NewCommitment, Nullified } from "../../generated/HinkalPool/HinkalPool";
+import {
+  HinkalProtocolStats,
+  HinkalOperationStats,
+  HinkalShieldERC20TokenStats,
+  HinkalUnshieldERC20TokenStats,
+  HinkalTransactTokenStats,
+  HinkalUnshieldNativeTokenStats,
+  HinkalShieldNativeTokenStats,
+  HinkalShieldNative,
+  HinkalUnshieldNative,
+  HinkalUnshieldERC20,
+  HinkalShieldERC20,
+  HinkalTransact,
+} from "../../generated/schema";
+
+import { getOperationFromLogs, Operations, parseTransferEvent } from "./utils";
+
+function createOrLoadProtocolStats(): HinkalProtocolStats {
+  const id = "hinkal-protocol-stats";
+  let stats = HinkalProtocolStats.load(id);
+
+  if (stats == null) {
+    stats = new HinkalProtocolStats(id);
+
+    stats.totalTxCount = BigInt.zero();
+    stats.totalGasUsed = BigInt.zero();
+
+    const shieldERC20 = new HinkalOperationStats(`${id}-shield-erc20`);
+    const unshieldERC20 = new HinkalOperationStats(`${id}-unshield-erc20`);
+    const shieldNative = new HinkalOperationStats(`${id}-shield-native`);
+    const unshieldNative = new HinkalOperationStats(`${id}-unshield-native`);
+    const transact = new HinkalOperationStats(`${id}-transact`);
+
+    shieldERC20.totalCount = BigInt.zero();
+    shieldERC20.totalGasUsed = BigInt.zero();
+
+    unshieldERC20.totalCount = BigInt.zero();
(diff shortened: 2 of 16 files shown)


subgraph/schemas/railgun.graphql
```diff
@@ -43,40 +43,13 @@ type RailgunShield @entity(immutable: true) {
   startPosition: BigInt!
   fees: [BigInt!]!
 
-  commitments: [RailgunCommitment!]! @derivedFrom(field: "shield")
```
NicoSerranoP, line 46: Why are we deleting commitments and ciphertexts? 🤔

subgraph/src/hinkal/utils.ts
```diff
@@ -0,0 +1,106 @@
+import { Address, BigInt, Bytes, ethereum } from "@graphprotocol/graph-ts";
+
+import { HinkalTransferLog } from "../../generated/schema";
+
+export const TRANSFER_TOPIC = Bytes.fromHexString("0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef");
```
NicoSerranoP, line 5: Non-blocking: what do you think about using the readable "Transfer(from,to,amount)" and then hashing it? That way it would be easier to read than the hash itself

subgraph/src/hinkal/utils.ts
```diff
@@ -0,0 +1,106 @@
...
+      if (log.topics[0].equals(TRANSFER_TOPIC)) {
+        events.push("TRANSFER");
+      }
+    }
+  }
+
+  return events;
+}
+
+export function parseTransferEvent(event: ethereum.Event, log: ethereum.Log, index: BigInt): HinkalTransferLog | null {
```
NicoSerranoP, line 80: Non-blocking suggestion: I like this function! It could be useful in other protocol benchmarks as well. When we get time we should try to generalize it (the return object at the moment returns a `HinkalTransferLog` but it should be something like `ERC20TransferLog`

subgraph/src/railgun/railgun-smart-wallet.ts
```diff
@@ -129,19 +126,7 @@ export function handleShield(event: ShieldEvent): void {
   shield.gasPrice = event.transaction.gasPrice;
 
   for (let index = 0; index < event.params.commitments.length; index += 1) {
-    const commitmentId = `${id}-c-${index.toString()}`;
```
NicoSerranoP, line 132: Why did you decided to remove commitments and ciphertexts? 😊

subgraph/src/hinkal/hinkal-pool.ts
```diff
@@ -0,0 +1,357 @@
...
+  const stats = createOrLoadProtocolStats();
+
+  stats.totalTxCount = stats.totalTxCount.plus(BigInt.fromI32(1));
+  stats.totalGasUsed = stats.totalGasUsed.plus(event.receipt!.gasUsed);
+
+  switch (operation) {
+    case Operations.UNSHIELD_ERC20: {
+      const unshield = new HinkalUnshieldERC20(event.transaction.hash.toHex());
+      const unshieldStats = HinkalOperationStats.load(stats.unshieldERC20);
+      const feeLog = parseTransferEvent(event, logs[0], BigInt.fromI32(0));
```
NicoSerranoP, line 291: Observation: we should benchmark fees as well? This is out of the scope but I feel it would be super useful as a user:

1. how much gas does it cost me?
2. how much fees to the protocol does it cost me?

NicoSerranoP (approved):
I left a couple of questions but I am happy with the PR! Feel free to reply or merge it

subgraph/src/hinkal/utils.ts
0xmad: There is a name `TRANSFER_TOPIC`, I don't think you need to not use hash.

subgraph/schemas/railgun.graphql
0xmad: No need to have them for now, so can be deleted

subgraph/src/hinkal/utils.ts
0xmad: It can be moved later to utils

subgraph/src/railgun/railgun-smart-wallet.ts
0xmad: No need to have them now

subgraph/src/hinkal/hinkal-pool.ts
0xmad: No, if there is a fee for relayer I just add it (pretty easy to add and can be added later to properties)

subgraph/src/hinkal/hinkal-pool.ts
```diff
@@ -0,0 +1,357 @@
...
+  const operation = getOperationFromLogs(logs);
+
+  if (![Operations.SHIELD_ERC20, Operations.INTERNAL_TRANSFER, Operations.SHIELD_NATIVE].includes(operation)) {
+    return;
+  }
+
+  const stats = createOrLoadProtocolStats();
+
+  stats.totalTxCount = stats.totalTxCount.plus(BigInt.fromI32(1));
+  stats.totalGasUsed = stats.totalGasUsed.plus(event.receipt!.gasUsed);
```
JohnGuilding, line 171: My understanding is that `handleNewCommitment` fires per `NewCommitment`. Does that not mean that txs that emit multiple `NewCommitment`'s would mean that this handler gets fired n times for a single transaction? And then `totalTxCount` and `totalGasUsed` would be getting double counted?

Should `totalTxCount` and `totalGasUsed` only get incremented per tx rather than per event?

And then follow up question would this be an issue with other benchmarks as well?

subgraph/src/hinkal/utils.ts
```diff
@@ -0,0 +1,106 @@
...
+  INTERNAL_TRANSFER,
+  SHIELD_NATIVE,
+  UNSHIELD_NATIVE,
+  UNKNOWN,
+}
+
+export function getOperationFromLogs(logs: ethereum.Log[]): Operations {
+  const events = extractEvents(logs);
+
+  const isUnshieldERC20 = events.includes("TRANSFER") && events.includes("NULLIFIED") && !events.includes("COMMITMENT");
```
JohnGuilding, line 25: Edge case question: Wondering if it possible that a transaction contains `Transfer` events from other unrelated contracts (multisend?, MEV?). Could this skew results by counting wrong events?

If so, can/should we only extract events emitted at the specific address we want

subgraph/tests/hinkal/hinkal-shield-erc20.test.ts
JohnGuilding, line 1: Missing test file for transfer?

subgraph/src/hinkal/hinkal-pool.ts
0xmad: It's handled by parsing event logs

subgraph/src/hinkal/utils.ts
0xmad: It's not possible because there is no event listeners for unrelated contracts.

subgraph/tests/hinkal/hinkal-shield-erc20.test.ts
0xmad: Will add with privacy pools

subgraph/src/hinkal/utils.ts
JohnGuilding: ok got it

subgraph/tests/hinkal/hinkal-shield-erc20.test.ts
JohnGuilding: ok cool

subgraph/src/hinkal/hinkal-pool.ts
JohnGuilding: Can you explain in more detail? I'm not sure I understand the logic you are referring to

0xmad: We have a specific combination of events and according to this we define operation. For instance, if we have `Transfer` and `Nullified` but don't have `NewCommitment` - it would be unshield for erc20.
This logic is implemented here: https://github.com/privacy-ethereum/private-transfers-benchmarks/blob/16c678694553e3db98b2efea25edc1bc1b2435c0/subgraph/src/hinkal/utils.ts

0xmad: And to prevent double counting there are some checks:
1. For `handleNewCommitment`:

```
if (![Operations.SHIELD_ERC20, Operations.INTERNAL_TRANSFER, Operations.SHIELD_NATIVE].includes(operation)) {
    return;
  }
```

2. For `handleNullified`:

```
if (![Operations.UNSHIELD_ERC20, Operations.UNSHIELD_NATIVE].includes(operation)) {
    return;
  }
```

So there is no overlap for events and if it happens it will be handled only by one handler, not both.

JohnGuilding: ok thanks for explanation

PR #200, opened 2026-05-13
feat(gas-benchmarks): add subgraph service with caching decorator

0xmad:
# What this PR does

Add subgraph service with caching decorator

## Confirmation

> [!IMPORTANT]

### DIFF
--- a/gas-benchmarks/src/index.ts
+++ b/gas-benchmarks/src/index.ts
@@ -4,6 +4,7 @@ import { Intmax } from "./intmax/index.js";
 import { Monero } from "./monero/index.js";
 import { PrivacyPools } from "./privacy-pools/index.js";
 import { Railgun } from "./railgun/index.js";
+// import { SubgraphService } from "./subgraph/index.js";
 import { TornadoCash } from "./tornado-cash/index.js";
 import { db } from "./utils/db.js";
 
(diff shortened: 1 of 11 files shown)


NicoSerranoP (approved):
Looks good to me 🚀 . One question: implementing a class decorator is not an overkill? Where should we be using the decorator apart from the `fetchRootQueryWithCache()` inside the `SubgraphService` class?

gas-benchmarks/src/index.ts
```diff
@@ -19,8 +20,9 @@ await db.read();
 
 const start = Date.now();
 
-// const client = new GraphQLClient(SUBGRAPH_URL);
-// const result = await client.request<TRootQuery>(RootQuery);
+// const subgraphService = await SubgraphService.getInstance();
+// const result = await subgraphService.fetchRootQueryWithCache();
```
NicoSerranoP, line 24: On each benchmark run, we will call `subgraphService.fetchRootQueryWithCache()` once and extract the protocol stats of all currently benchmarked protocols right? 

Is there a case where we will additionally fetch data from TheGraph? Or would it only be one call for all protocol stats?

0xmad: 1. Yes, all the benchmarks will be available in this root query and it's possible to call it from any benchmark because it's a singleton and results are cached. However, it's possible to call it once in entrypoint and pass only needed data.
2. As far as I remember, there should be some filter (latest transaction for 7 days and etc) or protocol specific transactions. For the filter logic we can just fetch the transactions and calculate the benchmarks manually for the current period. So there is a need in decorator because it prevents adding the same caching logic for different functions.

PR #225, opened 2026-05-25
feat(gas-benchmarks): add arbitrum subgraph support

0xmad:
# What this PR does

Add arbitrum subgraph support

## Confirmation

> [!IMPORTANT]


subgraph/schema.graphql
```diff
@@ -525,114 +525,6 @@ type TornadoCashWithdrawal @entity(immutable: true) {
   gasPrice: BigInt!
 }
 
-type VeilCashProtocolStats @entity(immutable: false) {
```
JohnGuilding, line 528: Might be lacking context from other PRs, but why is this schema removed?

gas-benchmarks/src/subgraph/curvy.ts
```diff
@@ -0,0 +1,17 @@
+import { graphql } from "../generated/mainnet/gql.js";
```
JohnGuilding: Should it be `arbitrum/gql.js`? Or is it meant to be mainnet?

JohnGuilding (approved):
Thanks! Nice clean PR

Does Arb need adding to this logic in `rpc.ts` or is this not needed anymore:
```ts
/** Pre-configured RPC clients keyed by chain ID */
const clients: Record<number, PublicClient | undefined> = {
  [mainnet.id]: createPublicClient({ chain: mainnet, transport: http(ETH_RPC_URL, { batch: true }) }),
  [scroll.id]: createPublicClient({ chain: scroll, transport: http(SCROLL_RPC_URL, { batch: true }) }),
  [sepolia.id]: createPublicClient({ chain: sepolia, transport: http(SEPOLIA_RPC_URL, { batch: true }) }),
};
```

subgraph/schema.graphql
0xmad: There is no veil cash added. It was added with Nico's PR and after codegen it's removed back.

0xmad:
@JohnGuilding there is no need to have arbitrum rpc endpoint because we are using subgraph.

PR #237, opened 2026-05-28
feat: add intmax subgraph support

0xmad:
# What this PR does

- [x] Support intmax deposit for mainnet
- [x] Support intmax withdrawal for scroll
- [x] Remove unused code

## Confirmation

> [!IMPORTANT]


subgraph/src/intmax/withdrawal.ts
```diff
@@ -0,0 +1,110 @@
...
+
+import {
+  IntmaxScrollWithdrawal,
+  IntmaxScrollWithdrawalStats,
+  IntmaxScrollOperationStats,
+  IntmaxScrollProtocolStats,
+} from "../../generated/schema";
+import { DirectWithdrawalQueued } from "../../generated/Withdrawal/Withdrawal";
+
+function createOrLoadProtocolStats(): IntmaxScrollProtocolStats {
```
NicoSerranoP, line 11: What if we reuse the `createOrLoadProtocolStats` from the other file to not duplicate code and have a single source of truth in case we need to modify something?

subgraph/src/intmax/withdrawal.ts
```diff
@@ -0,0 +1,110 @@
...
+      withdrawalStats.totalGasUsed = withdrawalStats.totalGasUsed.plus(event.receipt!.gasUsed);
+    }
+
+    withdrawalStats.save();
+  }
+
+  stats.save();
+
+  const withdrawalData = event.params.withdrawal;
+  const tokenIndex = withdrawalData.tokenIndex;
```
NicoSerranoP, line 105: There is no way to look up for the token address in Scroll as we did in the liquidityV2 handler? 😊 If not could we save the tokenIndex in the liquidityV2 handler so we can know which withdraws are native ETH and which are ERC20?

subgraph/src/intmax/withdrawal.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  if (withdrawalStats !== null) {
+    const tokenStats = createOrLoadWithdrawalStats(tokenIndex, withdrawalStats.id);
+
+    tokenStats.totalCount = tokenStats.totalCount.plus(BigInt.fromI32(1));
+
+    tokenStats.totalValue = tokenStats.totalValue.plus(withdrawalData.amount);
+
+    if (event.receipt !== null) {
+      tokenStats.totalGasUsed = tokenStats.totalGasUsed.plus(event.receipt!.gasUsed);
```
NicoSerranoP: `event.receipt!.gasUsed` is the gas used for the whole tx. These withdraw txs usually have more than one `DirectWithdrawalQueued` because withdrawals can be done in batches. The current logic will increment totalGasUsed per event rather than per tx. For example:

- 1 txs with 8 `DirectWithdrawalQueued` can cost N gas units.
- `totalGasUsed` = 8 x N
- In reality `totalGasUsed` is only N


I think we could solve this by looking at the event logs and counting how many `DirectWithdrawalQueued` are present in the tx. Then just adding (gasUnits / totalEvents). That way we would have a better approximate

subgraph/src/intmax/withdrawal.ts
0xmad: Protocol stats on mainnet and scroll are different ones. It's not possible to mix them.

subgraph/src/intmax/withdrawal.ts
0xmad: It's not possible to take data from mainnet subgraph to scroll. I can add token index for mainnet.

subgraph/src/intmax/withdrawal.ts
0xmad: Ok, will add it.

NicoSerranoP (approved):
🚀

PR #238, opened 2026-05-28
Add project evaluation for Aleo

AleoAlexander:
# What this PR does

Adds project evaluation for Aleo

## Confirmation

> [!IMPORTANT]

### DIFF
--- a/project-evaluations/src/data/evaluations/aleo.json
+++ b/project-evaluations/src/data/evaluations/aleo.json
@@ -1,10 +1,404 @@
 {
   "id": "aleo",
   "title": "Aleo",
-  "description": "Aleo is a permissionless private programability blockchain. It achieves private programability by using zero knowledge proofs to execute transactions and smart contract functions. Aleo has its own SnarkVM and the Leo programming language.",
-  "status": "pending",
-  "documentation": "https://developer.aleo.org/concepts/network/core_architecture",
-  "categories": ["Private L1"],
-  "properties": [],
-  "sourceUrls": []
+  "description": "Aleo is a Layer-1 blockchain designed for private, programmable applications. It uses zero-knowledge proofs to let developers build apps where user data stays private, while still allowing the rest of the network to verify that everything is correct. Aleo has its own zkVM (snarkVM) and custom programming language (Leo).",
+  "status": "complete",
+  "documentation": "https://docs.aleo.org",
+  "categories": ["Private L1", "Shielded Pool", "Zero Knowledge Proofs (ZKPs)"],
+  "properties": [
+    {
+      "name": "Anonymity set size",
+      "value": ""
+    },
+    {
+      "name": "Confidentiality",
+      "value": "Yes",
+      "notes": "Aleo provides full confidentiality for all addresses, where both balances and transaction amounts remain private through zero-knowledge proofs.  A single Aleo addresses can hold both shielded and unshielded funds.  Shielded transactions keeps the addresses and transaction data shielded from the public. Unshielded transactions work similarly to Ethereum and do not offer privacy, with addresses and values publicly visible.  Users can convert between shielded/unshielded funds at any time.",
+      "citations": [
+        {
+          "cited_text": "`transfer_private`: Fully private — sender, receiver, and amount are all hidden. `transfer_private` consumes a record from the sender and produces two new records: one for the recipient and one for the sender's change. No public information is revealed.",
+          "source": "https://docs.aleo.org/learn/core-concepts/credits-and-transfers"
+        },
+        {
+          "cited_text": "Records are Aleo's approach to private state. They are units of encrypted data owned by a specific user. Records work similarly to Bitcoin's UTXO model; when you \"spend\" a record, it's consumed and a new one is created.  Only the owner (and anyone with their view key) can decrypt a record's contents. Records are stored onchain in encrypted form, [and] executed and produced offchain as part of a transition, with a ZK proof submitted for verification",
+          "source": "https://docs.aleo.org/learn/core-concepts/public-and-private-state"
+        }
+      ]
+    },
+    {
+      "name": "Anonymity",
+      "value": "Yes",
+      "notes": "Aleo supports full anonymity for both sender and receiver when performing shielded transactions. A shielded transaction keeps the addresses and transaction amounts shielded from the public using zero-knowledge proofs to allow transaction data to be encrypted but remain verifiable by network nodes.  While Aleo also supports unshielded transactions that work like Ethereum with public visibility, the availability of fully shielded transactions means the protocol supports sender and receiver anonymity relative to any uninvolved parties.",
+      "citations": [
+        {
+          "cited_text": "`transfer_private`: Fully private — sender, receiver, and amount are all hidden. `transfer_private` consumes a record from the sender and produces two new records: one for the recipient and one for the sender's change. No public information is revealed.",
+          "source": "https://docs.aleo.org/learn/core-concepts/credits-and-transfers"
+        },
+        {
+          "cited_text": "Records also include an encrypted sender_ciphertext field (as of Consensus V8), which stores the address of the transaction signer who created the record. This is encrypted alongside the record data, so only the record owner can see who sent it. This allows recipients to identify the sender of a private transfer without that information being publicly visible onchain.",
+          "source": "https://docs.aleo.org/learn/core-concepts/public-and-private-state"
+        }
+      ]
+    },
+    {
+      "name": "Asset privacy",
+      "value": "No",
+      "notes": "Aleo supports several token types, including its native asset Aleo Credits (ALEO), the USDCx and USAD private stablecoins, and a variety of wrapped assets (ETH, SOL, wBTC, etc.).  However, these tokens are all defined by visible smart contracts (programs) onchain, so function execution and consequently asset type are fully visible. ",
+      "citations": [
+        {
+          "cited_text": "Aleo Credits are the native token of the Aleo network. They're used for transaction fees, staking, and transferring value. All credit operations are handled by the built-in credits.aleo program.",
+          "source": "https://docs.aleo.org/learn/core-concepts/credits-and-transfers"
+        },
+        {
+          "cited_text": "https://explorer.provable.com/program/usad_stablecoin.aleo , https://explorer.provable.com/program/usdcx_stablecoin.aleo",
+          "source": "https://docs.aleo.org/build/common-uses/private_stablecoins"
+        }
+      ]
+    },
+    {
+      "name": "Plausible deniability",
+      "value": "No",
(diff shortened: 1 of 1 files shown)


project-evaluations/src/data/evaluations/aleo.json
```diff
@@ -1,10 +1,403 @@
...
+    },
+    {
+      "name": "Escape hatch",
+      "value": "N/A",
+      "notes": "Aleo is a standalone Layer 1 blockchain where assets can be sent both shielded and unshielded through standard onchain transactions.  There is no separate system from which users need to 'exit' - users can move native or ARC-20 funds at any time.  The exception to this rule is for ARC-22 (Compliant Token Standard) tokens such as USDCx and USAD.  These tokens enforce a privacy-preserving freezelist mechanism, which prevents blocked addresses from sending, receiving, shielding, or unshielding any tokens. There is no escape hatch or ragequit functionality for blocked addresses. This does not apply to the native asset ALEO.",
+      "needsResearchReview": "Needs explicit source once ARC-20 and ARC-22 standards are finalized"
+    },
+    {
+      "name": "Upgradeability",
+      "value": ["Network upgrade (hard/soft fork)", "Single admin", "Multi-sig", "DAO", "Immutable"],
```
NicoSerranoP:
```suggestion
      "value": ["Network upgrade (hard/soft fork)"],
```

NicoSerranoP: I would only keep the Network upgrade option because we are talking about ALEO and its native token (credit). I understand that deployed smart contracts can have their own governance methods but lets keep the scope to the network native token (in this case Aleo credits)

project-evaluations/src/data/evaluations/aleo.json
```diff
@@ -1,10 +1,403 @@
...
+        },
+        {
+          "cited_text": "Source code",
+          "source": "https://github.com/ProvableHQ/snarkVM/blob/1c7bce51fdeb3a27662c54cca93de448ab729a94/console/network/src/consensus_heights.rs#L129"
+        }
+      ]
+    },
+    {
+      "name": "Client-side proving",
+      "value": "Yes",
```
NicoSerranoP: Where is the proof generated when I perform a private transfer in the default Aleo's wallet? We need to assign the properties values assuming the default scenario that most users will be.

If you can theoretically have client-side proving but in practice it is mostly relayed to an external prover then the value will be "No".

Also if the user is required to download another app (prover) and connect it to the Aleo's wallet in order to enable client-side proving, then the value will be "No". Client-side proving should be the default path

project-evaluations/src/data/evaluations/aleo.json
```diff
@@ -1,10 +1,403 @@
...
+        },
+        {
+          "cited_text": "Generating zero-knowledge proofs is computationally intensive. Delegated proving allows users to offload proof generation to a third-party prover — without giving up control of their funds or exposing their private key. ",
+          "source": "https://docs.aleo.org/learn/advanced/delegated-proving"
+        }
+      ]
+    },
+    {
+      "name": "Third-party inspectability",
+      "value": "No",
```
NicoSerranoP, line 185: If client-side proving is set to "No" then Third-part inspectability will be "Yes" because the external prover could check the users private data and leak it

project-evaluations/src/data/evaluations/aleo.json
```diff
@@ -1,10 +1,403 @@
...
+        },
+        {
+          "cited_text": "Addresses on the freeze list are blocked from all public transfers. Private transfers enforce the same restriction without revealing the sender's address: the sender supplies a Merkle non-membership proof demonstrating their address is not a leaf in the freeze list tree.  Every operation that crosses the public/private boundary (mint_private, burn_private, and transfer_public_to_private) emits a ComplianceRecord to a designated compliance officer address. This gives the issuer an audit trail for regulatory purposes without exposing information publicly onchain.",
+          "source": "https://docs.aleo.org/build/common-uses/private_stablecoins"
+        }
+      ]
+    },
+    {
+      "name": "Enforcement entities",
+      "value": "Admin",
```
NicoSerranoP:
```suggestion
      "value": "None",
```

NicoSerranoP: Lets focus on the native token private transfers. In that case we do not have a enforcement entity

project-evaluations/src/data/evaluations/aleo.json
```diff
@@ -1,10 +1,403 @@
...
+        },
+        {
+          "cited_text": "Addresses on the freeze list are blocked from all public transfers. Private transfers enforce the same restriction without revealing the sender's address: the sender supplies a Merkle non-membership proof demonstrating their address is not a leaf in the freeze list tree.  Every operation that crosses the public/private boundary (mint_private, burn_private, and transfer_public_to_private) emits a ComplianceRecord to a designated compliance officer address. This gives the issuer an audit trail for regulatory purposes without exposing information publicly onchain.",
+          "source": "https://docs.aleo.org/build/common-uses/private_stablecoins"
+        }
+      ]
+    },
+    {
+      "name": "Point of enforcement",
+      "value": ["Deposit", "Transfer", "Withdrawal"],
```
NicoSerranoP:
```suggestion
      "value": ["None"],
```

NicoSerranoP: Let's focus only on the native token private transfers

project-evaluations/src/data/evaluations/aleo.json
```diff
@@ -1,10 +1,403 @@
...
+      "citations": [
+        {
+          "cited_text": "Every operation that crosses the public/private boundary (mint_private, burn_private, and transfer_public_to_private) emits a ComplianceRecord to a designated compliance officer address. This gives the issuer an audit trail for regulatory purposes without exposing information publicly onchain.",
+          "source": "https://docs.aleo.org/build/common-uses/private_stablecoins"
+        }
+      ]
+    },
+    {
+      "name": "Selective disclosure: viewing entity",
+      "value": ["User"],
```
NicoSerranoP:
```suggestion
      "value": ["User", "Voluntary third-party"],
```

NicoSerranoP: Users can share their viewing key voluntarily to third-parties to check their txs

project-evaluations/src/data/evaluations/aleo.json
```diff
@@ -1,10 +1,403 @@
...
+          "source": "https://docs.aleo.org/learn/core-concepts/accounts-and-keys"
+        }
+      ]
+    },
+    {
+      "name": "Selective disclosure: viewing control",
+      "value": "Pre-defined",
+      "notes": "Aleo implements pre-defined selective disclosure viewing control through the view key mechanism.Addresses can share transaction details with trusted third parties via a view key, which grants read access but not spend authority, enabling selective disclosure where transactions are auditable but disclosure is under the participant's control, allowing compliance with payment auditing, tax regulations, or anti-money laundering rules. The view key mechanism is a fixed feature of the protocol rather than a programmable system with conditional logic.",
```
NicoSerranoP:
```suggestion
      "notes": "Aleo implements pre-defined selective disclosure viewing control through the view key mechanism. Addresses can share transaction details with trusted third parties via a view key, which grants read access but not spend authority, enabling selective disclosure where transactions are auditable but disclosure is under the participant's control, allowing compliance with payment auditing, tax regulations, or anti-money laundering rules. The view key mechanism is a fixed feature of the protocol rather than a programmable system with conditional logic.",
```

project-evaluations/src/data/schema.ts
```diff
@@ -168,7 +168,7 @@ export const PROPERTY_DEFINITIONS: PropertyContent[] = [
     group: "Decentralization & Security",
     description: "How upgrades to the system are performed.",
     metric: "Single admin / Multi-sig / DAO / Network upgrade (hard/soft fork) / Immutable",
-    inputType: "select",
+    inputType: "multi-select",
```
NicoSerranoP: I think we can avoid this change if we address one of the comments above

NicoSerranoP (changes requested):
@AleoAlexander thank you for your contribution! 🥳  I left some comments and suggestions. If you need we can have a async discussion on each comment in order to resolve them

zklimaleo:
Thanks for the thorough review! Agree with all comments, except **Third-party inspectability** because delegated proving is encrypted, so the prover never sees the user's private inputs. Added clarification in notes. The privacy level also stated in default Aleo wallet, Shield's setting, as shown in the screenshot.
Screenshot 2026-06-12 at 11 18 40 PM

@NicoSerranoP

NicoSerranoP:
Super interesting work around delegated proving. I could not find to much info in the Aleo Docs website: https://docs.aleo.org/learn/advanced/delegated-proving . The docs even states that:

What the Prover Sees
- All input values in plaintext

After your comment I read more about Shield and the encrypted delegate proving using TEEs + external provers is a super interesting solution! The main reference for this is: https://docs.provable.com/docs/api/services/post-prove-encrypted


I think it makes sense to leave "Third-party inspectability": "Yes", but we would need a bit more explanation covering:

- that Shield is the default Aleo's wallet. In the [documentation wallet section](https://docs.aleo.org/participate/wallets)
- how the encrypted delegated proving mechanism works. I could not find any technical details about it except from this [page](https://docs.provable.com/docs/api/services/post-prove-encrypted). If you have an open source documentation it would be great. Remember to use the "citations" field to add these links


Thanks for the changes! I have resolved the conversations. Let's add this explainer and we will be ready to merge it! 🚀

zklimaleo:
Thanks for the quick response! Apologies that the docs is not up to date, the integration guide did mentioned more about encrypted delegated proving flow https://docs.aleo.org/build/sdk/guides/delegate_proving#flow-encrypted-proving and the reference you found is the correct API too. I will need to improve the docs better.

While it is not explicitly mentioned, Shield wallet is actually developed by the Aleo labs company, Provable because we want to maintain fairness across similar wallet products. It is the most actively maintained wallet in the ecosystem that comes with many extra features that other wallets might not have, such as privacy-preserving delegated proving service, privacy-preserving record scanning service, cross-chain swaps, on / off ramps etc.. That's why it is the most recommended "default" wallet right now for Aleo users. I can update the documentation to include a recommendation based on the mentioned features so that it can be cited.

@NicoSerranoP

NicoSerranoP:
@zklimaleo your welcome :) please help us out with that explaination in the evaluation file and we will be ready to merge it 🚀 . Remember that we need this PR rebased to latest main and also 1 commit per PR. To squash your commits to a single one follow these steps:

```bash
git checkout main
git pull
git checkout your-branch
git rebase main
git reset $(git merge-base main $(git rev-parse --abbrev-ref HEAD))
```

zklimaleo:
@NicoSerranoP Added mention of Shield wallet in Client-side proving & Third-party inspectability notes and squashed commits into 1

NicoSerranoP (approved):

zklimaleo:
Seems like description text is too long

PR #251, opened 2026-06-06
Feat/eval client indexing

JohnGuilding:
# What this PR does

- reduce client-side indexing to binary choice
- update skill
- anoma changed to no scanning since third party indexer does the scanning

Partial scanning is currently ambiguous and could capture multiple axes: it could capture device type, or performance of scanning. This change simplifies this property by only having scanning and not scanning as options

Ways we can break this down:

1. Binary choice (this PR): Scanning / No Scanning, explain nuance in the note. Simple, but doesn't capture all nuance

2. Three options based on who scans — Device / Delegated / No scan. Captures which device + trust boundary (delegated = a third party indexer scans)

3. Three options based on how much is scanned — Full / Filtered / No scan. Captures the full scan vs view-tag differences which impacts performance

To avoid a more significant re-analysis at this stage when don't have much time, I opted to use the binary choice. We can revisit this in the future if we want to

## Confirmation

> [!IMPORTANT]

### DIFF
--- a/project-evaluations/.claude/skills/property-rules-state-model/SKILL.md
+++ b/project-evaluations/.claude/skills/property-rules-state-model/SKILL.md
@@ -21,7 +21,12 @@ For services with no protocol contracts at all — off-chain routing aggregators
 
 ## Client-side indexing
 
-(no property-specific rule yet — apply cross-cutting rules)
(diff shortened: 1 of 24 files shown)


NicoSerranoP (approved):

NicoSerranoP:
We need to merge https://github.com/privacy-ethereum/private-transfers-benchmarks/pull/254 first right?

PR #254, opened 2026-06-06
Feat/eval state scalability

JohnGuilding:
# What this PR does

- It is not completely clear what infinity grow and temporal grow mean from descriptions. We also haven't used temporal grow so far 
- update to more specific naming: per transaction, per account
- Fluidkey change to per transactions since new accounts are deployed
- Redact and Tongo updates to per account. Stateless implies not state
- Scroll cloak is an EVM but hidden by access control and execution is proved by zkevm, and thus state changes are per account assuming an erc20 model
- additional tightening is to refer to this state as onchain state. This makes protocols like stateless intmax make sense, as intmax is not stateless, the state is just client side and not onchain. Same with houdini swap, state exists in backend infra, but nothing exists onchain from my understanding 

## Confirmation

> [!IMPORTANT]


project-evaluations/src/data/evaluations/scroll-cloak.json
```diff
@@ -427,8 +427,8 @@
     },
     {
       "name": "Private State Scalability",
-      "value": "Infinity grow",
-      "notes": "Each Cloak instance keeps its own isolated account-based EVM ledger, so state accumulates per transaction as on any other EVM chain, growing monotonically as commitments and account balances are added over the life of the chain.",
+      "value": "Per Account",
```
NicoSerranoP, line 430: Cloak could be considered "Stateless" due to its storage been kept in the sequencer server (third party dependency). The on-chain state does not grow, Cloak instances only pushes commitments to Scroll. All the accounts and tx data is kept in the sequencer disk and not on-chain

NicoSerranoP (approved):
@JohnGuilding just one comment about Cloak :)

project-evaluations/src/data/evaluations/scroll-cloak.json
JohnGuilding: I think I agree - want to compare to more protocols and then will address. Issue [here](https://github.com/privacy-ethereum/private-transfers-benchmarks/issues/264)

PR #255, opened 2026-06-06
Feat/eval time display

JohnGuilding:
# What this PR does

- normalise time values so all values are in seconds and get formatted to a human readable version. 3600 seconds is not immediately readable but 1 hour is.
- fix zcash bugs raised by zcash contributor feedback in W3PN Telegram
- update intmax finality value to latest from l2beat

## Confirmation

> [!IMPORTANT]

### DIFF
--- a/project-evaluations/src/utils.ts
+++ b/project-evaluations/src/utils.ts
@@ -3,8 +3,28 @@ import type { Evaluation } from "./types.js";
 const NUMBER_OF_SECRETS_REGEX = /number of secrets/i;
 const PROJECT_README_PATH_REGEX = /subgraph\/src\/([^/]+)\/README\.md$/;
 
-/** Properties shown without a yes/no colour — the answer is contextual, not better or worse. */
-const NEUTRAL_TONE_PROPERTIES = new Set(["Plausible deniability"]);
+/** Formats seconds, or a "min-max" range, into readable parts, e.g. "9000" → "2 hours, 30 minutes", "15-30" → "15 seconds - 30 seconds". */
+function formatDuration(secondsValue: string): string {
+  const seconds = secondsValue.split("-").map(Number);
+  if (seconds.some(Number.isNaN)) return secondsValue;
+  return seconds.map(formatSeconds).join(" - ");
+}
+
+/** Formats a whole number of seconds into its non-zero parts, e.g. 9000 → "2 hours, 30 minutes". */
+function formatSeconds(totalSeconds: number): string {
+  const days = Math.floor(totalSeconds / 86400);
+  const hours = Math.floor((totalSeconds % 86400) / 3600);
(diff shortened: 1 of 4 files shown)


project-evaluations/src/utils.ts
```diff
@@ -3,8 +3,28 @@ import type { Evaluation } from "./types.js";
...
+
+/** Formats a whole number of seconds into its non-zero parts, e.g. 9000 → "2 hours, 30 minutes". */
+function formatSeconds(totalSeconds: number): string {
+  const days = Math.floor(totalSeconds / 86400);
+  const hours = Math.floor((totalSeconds % 86400) / 3600);
+  const minutes = Math.floor((totalSeconds % 3600) / 60);
+  const seconds = totalSeconds % 60;
+
+  const parts: string[] = [];
+  if (days > 0) parts.push(`${days} day${days === 1 ? "" : "s"}`);
```
NicoSerranoP, line 21: Nit pick: should we use days? I would think that the text in the UI will not look great with a long text: "2 days 3 hours 45 seconds"

Just a comment 😊 . In practice I am not 100% sure if we have a protocol that takes more than 1 day to finalize

project-evaluations/src/utils.ts
```diff
@@ -48,7 +76,8 @@ type CellTone = "yes" | "no" | "warn" | "";
  * @dev it is currently used to mark as yellow if number of secrets is 2 or greater
  */
 export function cellTone({ value, propertyName }: ICellToneParams): CellTone {
-  if (NEUTRAL_TONE_PROPERTIES.has(propertyName)) {
+  const neutralToneProperties = new Set(["Plausible deniability"]);
```
NicoSerranoP, line 79: I liked the `const NEUTRAL_TONE_PROPERTIES = new Set(["Plausible deniability"]);` on top of the file because I feel it allows us to add new neutral tone properties in the future. I feel it is easier to access and does not initializes every time the function `cellTone` is called

NicoSerranoP (approved):
Left 2 comments for nitpick observations 😊 .

project-evaluations/src/utils.ts
JohnGuilding: I'm unsure these points mean it needs to be a top level const. We can still add new neutral tones by expanding the set inline, and the performance difference is negligible.  

I think it harmed readability unnecessarily by separating logic from the function. It didn't need to be abstracted — we can abstract it again if we ever need to use it anywhere else.

project-evaluations/src/utils.ts
JohnGuilding: Yeah I'm open to that. Lets see what it looks like in UI and then edit if it looks bad

project-evaluations/src/utils.ts
NicoSerranoP: Agree :) we can abstract it later if required

Issue #273, opened 2026-06-16
fix(benchmarks): check average gas used in privacy pools


NicoSerranoP:
## Description

A [deposit tx](https://etherscan.io/tx/0x850683b8268f20e0b4fc0fcbe538f4b6eef2e755fffb4a93b18a7add717b461a) in Privacy Pools costs $0.14 (0.00007728015018456 ETH). The data gathered in the benchmarks shows ~$3. We need to investigate why it is happening

## Acceptance Criteria

- Average gas costs similar to real txs of Privacy Pools

PR #274, opened 2026-06-16
feat(evaluation): add correct benchmarks to dashboard

### LINKED ISSUE #261: fix(subgraph): get correct benchmark gas data
## Description

It looks like our benchmarks are returning wrong data. For example a Railgun or Fluidkey transaction is around 0.01 ($10) to 0.03 ($30). These are not real values. We need to find the bug and fix it. Some iterations:

1. The benchmarks are historical so that could be causing the differences. I deployed the mainnet subgraph (v0.0.52) to a 3 months old block but we were still getting similar values
2. In Railgun we are considering every cyphertext as a new count but we are summing the total gas cost used in the transaction
3. Some transactions could have been bundled with other txs so the gas used might be higher
## Acceptance Criteria

- Add acceptance criteria here
Get real benchmark data when running `pnpm run benchmark`

NicoSerranoP:
# What this PR does

- Code restructure: move constants and enums to their own respective files

- Gets the correct average gas used cost for all protocols. Fixes: https://github.com/privacy-ethereum/private-transfers-benchmarks/issues/261 . It looks like Privacy Pools has a higher average that real txs, I created an issue and I will investigate further: https://github.com/privacy-ethereum/private-transfers-benchmarks/issues/273

- Gets gas prices for the different networks depending on the protocol

## Confirmation

> [!IMPORTANT]


project-evaluations/scripts/sync-benchmarks/constants.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export const FALLBACK_ETH_USD = 1800;
+
+// N gwei (not confuse with wei) represented as ETH
+export const FALLBACK_GAS_ETH_MAINNET = 3 * 1e-10;
+export const FALLBACK_GAS_ETH_ARBITRUM = 2 * 1e-11;
+export const FALLBACK_GAS_ETH_SEPOLIA = 1 * 1e-9;
+export const FALLBACK_GAS_ETH_BASE = 6 * 1e-12;
+export const FALLBACK_GAS_ETH_SCROLL = 1.2 * 1e-13;
+export const FALLBACK_GAS_ETH_STARKNET = 2 * 1e-10;
```
0xmad: Why do you need gwei? Subgraph returns wei and it's easier to calculate everything in wei using big int.

NicoSerranoP:
@codex

project-evaluations/scripts/sync-benchmarks/constants.ts
NicoSerranoP: Sorry that was a leftover comment. The fallback gas values are in wei. Subgraph and the RPCs return values in wei

project-evaluations/scripts/sync-benchmarks/constants.ts
```diff
@@ -0,0 +1,61 @@
+import { config } from "dotenv";
+
+config();
+
+if (!process.env.ETH_RPC_URL) {
+  throw new Error("ETH_RPC_URL is not set in the environment variables");
+}
```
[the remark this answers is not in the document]
NicoSerranoP: Do not worry. The fallback values will work in case the RPC does not respond

project-evaluations/scripts/sync-benchmarks/constants.ts
0xmad: No, I mean you have float number here. If you have calculations in wei you have integer (not float), it's safer for calculation (see how js works with float numbers here `0.1 + 0.1 + 0.1`)

NicoSerranoP: @0xmad I just made the changes. The gas units are computed as `totalGasUsed / totalCount`. I am truncating the integers because the values are high enough to loss a bit of precision (the cheapest one are 21,000 which will not be affected by the decimals). 

Can you review it again to merge it asap?

0xmad (approved):
@NicoSerranoP thanks!

project-evaluations/scripts/sync-benchmarks/constants.ts
```diff
@@ -0,0 +1,60 @@
...
+} = process.env;
+
+export const FALLBACK_ETH_USD = 1800;
+
+export const FALLBACK_GAS_ETH_MAINNET = BigInt(3e8);
+export const FALLBACK_GAS_ETH_ARBITRUM = BigInt(2e7);
+export const FALLBACK_GAS_ETH_SEPOLIA = BigInt(1e9);
+export const FALLBACK_GAS_ETH_BASE = BigInt(6e6);
+export const FALLBACK_GAS_ETH_SCROLL = BigInt(1.2e5);
+export const FALLBACK_GAS_ETH_STARKNET = 2n;
```
JohnGuilding, line 50: Why do we have fallback values? Would this not result in inaccurate calculations if they are used. Shouldn't we throw an exception if we can't fetch these values?

project-evaluations/scripts/sync-benchmarks/utils.ts
```diff
@@ -91,8 +86,13 @@ export async function fetchGasPrices(): Promise<IGasPrices> {
...
  */
-export function buildCostValueAndNotes(gasUnits: number, gasPrices: IGasPrices): IValueAndNotes {
-  const ethCost = gasUnits * gasPrices.gasPriceInETH;
+export function buildCostValueAndNotes(
+  gasUnits: number,
+  gasPrices: IGasPrices,
+  network: Network = Network.MAINNET,
+): IValueAndNotes {
+  const ethCostInWei = BigInt(Math.trunc(gasUnits)) * gasPrices.inETH[network];
+  const ethCost = Number(formatEther(ethCostInWei));
```
JohnGuilding, line 95: What was the fix in the end? Was it this logic?

JohnGuilding (commented):
Great to get this in 🎉

Two questions

PR #272, opened 2026-06-16
Feat/fairblock evaluation

JohnGuilding:
# What this PR does

- add fairblock evaluation
- merge https://github.com/privacy-ethereum/private-transfers-benchmarks/pull/271 first

## Confirmation

> [!IMPORTANT]
> We do not accept minor grammatical fixes (e.g., correcting typos, rewording sentences) unless they significantly improve clarity in technical documentation. These contributions, while appreciated, are not a priority for merging. If there is a grammatical error feel free to message the team.

- [x] I have supplied a PR description that explains what the PR does.
- [x] I have linked a github issue to the PR if one exists.
- [x] I ran and verified that all tests pass locally.

PR #275, opened 2026-06-17
Feat/merces evaluation

JohnGuilding:
# What this PR does

- add merces evaluation
- merge https://github.com/privacy-ethereum/private-transfers-benchmarks/pull/272 first

## Confirmation

> [!IMPORTANT]

### DIFF
--- a/project-evaluations/src/data/evaluations/merces.json
+++ b/project-evaluations/src/data/evaluations/merces.json
@@ -2,9 +2,522 @@
   "id": "merces",
   "title": "Merces",
   "description": "Merces by TACEO is an implementation of Private Shared State (PSS), providing a confidential stablecoin transfer system. It is deployed on Base and the Arc testnet.",
+  "documentation": "https://docs.taceo.io/docs/finance-solutions/overview/",
   "status": "pending",
-  "documentation": "https://merces-dashboard.taceo.io/base/introduction/overview",
-  "categories": ["CoSNARKs", "Zero Knowledge Proofs (ZKPs)", "Multi Party Computation (MPC)"],
-  "properties": [],
-  "sourceUrls": []
+  "categories": [
+    "CoSNARKs",
+    "Zero Knowledge Proofs (ZKPs)",
+    "Multi Party Computation (MPC)"
+  ],
+  "properties": [
+    {
+      "name": "Anonymity set size",
+      "value": ""
+    },
+    {
+      "name": "Confidentiality",
+      "value": "Yes",
+      "notes": "Balances are held as secret shares spread across the three-node TACEO MPC network, with state transitions verified on-chain via CoSNARKs, so no single party including TACEO can observe what an account holds. Transfer amounts are posted only as commitments alongside encrypted ciphertexts for the MPC nodes, with the plaintext value never exposed on-chain. Sender, receiver, and amount are each hidden behind commitments that the MPC-produced proof later verifies.",
+      "citations": [
+        {
+          "cited_text": "Balances are held as secret shares across the TACEO Network, state\ntransitions are verified onchain via CoSNARKs, and no single party (including TACEO) sees what\nanyone holds or sends. ",
+          "source": "https://docs.taceo.io/docs/finance-solutions/overview/"
+        },
+        {
+          "cited_text": "Since the sender, receiver and actual token amount is hidden, they are only given as a commitment, which later will be used to verify the ZK proof produced by the MPC network.",
+          "source": "https://core.taceo.io/articles/merces-deep-dive/"
+        }
+      ]
+    },
+    {
+      "name": "Anonymity",
+      "value": "Yes",
+      "notes": "Fully private transfers conceal sender, receiver, and amount, each supplied only as a commitment that the MPC-produced proof later verifies. Balances are encrypted, transfers hide all parties and amounts, and every state transition is checked by an on-chain zero-knowledge proof.",
+      "citations": [
+        {
+          "cited_text": "Since the sender, receiver and actual token amount is hidden, they are only given as a commitment, which later will be used to verify the ZK proof produced by the MPC network.",
+          "source": "https://core.taceo.io/articles/merces-deep-dive/"
+        },
+        {
+          "cited_text": "Balances are encrypted, transfers hide all parties and amounts, and every state transition is verified by a zero-knowledge proof onchain. ",
+          "source": "https://core.taceo.io/articles/merces-deep-dive/"
+        }
+      ]
+    },
+    {
+      "name": "Asset privacy",
+      "value": "No",
+      "notes": "Each deployment wraps one specific ERC-20, and the contract holds the actual tokens backing every private balance, so observers see which token is wrapped even though balances are encrypted. The encrypted Merkle tree tracks a single token type. Supporting multiple assets within the same encrypted tree is named as future work, not a live capability.",
+      "citations": [
+        {
+          "cited_text": "It holds the actual ERC-20 tokens backing all private balances, stores the current commitment (Merkle root) to the encrypted balance tree",
+          "source": "https://core.taceo.io/articles/merces-deep-dive/"
+        },
+        {
+          "cited_text": "supporting multiple assets within the same encrypted Merkle tree, enabling private interactions with external DeFi contracts while preserving the proof model",
+          "source": "https://core.taceo.io/articles/merces-deep-dive/"
+        }
+      ]
+    },
+    {
+      "name": "Plausible deniability",
+      "value": "No",
+      "notes": "Interacting with Merces is detectable on-chain. The entry path is a deposit registered in the contract's action queue rather than an ordinary token transfer, so the user posts commitments and encrypted data to a protocol-specific contract. That registration is visibly distinct from a standard ERC-20 transaction, marking the address as a Merces user.",
+      "citations": [
+        {
+          "cited_text": "stores the current commitment (Merkle root) to the encrypted balance tree, and maintains an action queue where users register intents: deposits, transfers, withdrawals.",
+          "source": "https://core.taceo.io/articles/merces-deep-dive/"
+        },
+        {
+          "cited_text": "A user registers an intent, such as a private transfer, by posting commitments and encrypted data onchain. ",
+          "source": "https://core.taceo.io/articles/merces-deep-dive/"
+        }
+      ]
+    },
+    {
+      "name": "On-chain gas cost: deposit",
+      "value": ""
+    },
+    {
+      "name": "On-chain gas cost: transfer",
+      "value": ""
+    },
+    {
+      "name": "On-chain gas cost: withdraw",
+      "value": ""
+    },
+    {
+      "name": "Time-to-finality",
+      "value": "Underlying chain",
+      "notes": "Merces runs as a standard Solidity contract on existing EVM chains with no settlement step of its own, so finality is inherited from the host chain. It is deployed on the Plasma testnet, with finality cadence set by whichever network hosts the contract rather than by any protocol-imposed waiting period.",
+      "citations": [
+        {
+          "cited_text": "The system runs as a standard Solidity contract, no protocol-level modifications, no custom chain, deployable wherever the EVM does.\n\n",
+          "source": "https://core.taceo.io/articles/merces-deep-dive/"
+        },
+        {
+          "cited_text": "The system is deployed on the Plasma testnet. ",
+          "source": "https://core.taceo.io/articles/merces-deep-dive/"
+        }
+      ]
+    },
+    {
+      "name": "Number of secrets",
+      "value": "1",
+      "notes": "The user stores one secret: the sender secret key for their private account, whose public key is registered in the on-chain ID registry. Spends are authorised by proving knowledge of this key against the registry root via a Merkle-tree proof during withdraw and transfer. Per-transfer encryption keypairs are sampled fresh through ECDH against the MPC nodes' public keys and are not persisted.",
+      "citations": [
+        {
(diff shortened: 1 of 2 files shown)


project-evaluations/src/data/evaluations/blanksquare.json
```diff
@@ -384,8 +384,8 @@
     },
     {
       "name": "Access to DeFi",
-      "value": "Access to external, but limited choice of DeFi protocols",
-      "notes": "Shielded assets are traded on public DeFi venues rather than a dedicated shielded DEX, so reaching protocols such as Uniswap or Cowswap is supported through the public route. A built-in Shielder-SDK integration with various DEXes is planned but not yet shipped, so today the path runs through the public withdraw route rather than a native in-pool DeFi runtime, which bounds what shielded balances can reach directly.",
+      "value": ["No access to DeFi"],
+      "notes": "Blanksqaure is advertised as a composable privacy stack for wallets and DeFi apps. As of writing, no DeFI integrations are live. Accessing DeFi means withdrawing from the shielded pool. A DEX integration is planned, but not yet live.",
```
NicoSerranoP:
```suggestion
      "notes": "Blanksquare is advertised as a composable privacy stack for wallets and DeFi apps. As of writing, no DeFI integrations are live. Accessing DeFi means withdrawing from the shielded pool. A DEX integration is planned, but not yet live.",
```

project-evaluations/src/data/evaluations/merces.json
```diff
@@ -1,10 +1,558 @@
...
+      "name": "On-chain gas cost: transfer",
+      "value": ""
+    },
+    {
+      "name": "On-chain gas cost: withdraw",
+      "value": ""
+    },
+    {
+      "name": "Time-to-finality",
+      "value": "Underlying chain",
```
NicoSerranoP, line 91: Comment: I think we should put the minimum external dependency time it takes between registering the intent and the external dependency performing the computation. In Merces a user post a operation intent (e.g. transfer) and the MPC network takes that intents and updates the shared private state. The same will happen with co-processor-based protocols like Zama, Redact, Soda Labs, etc.


The problem with this value is that documentation does not state the specific time to finality in most of the cases. I feel we could have a value similar to "Underlying chain + co-processor time" to state that it might take more time than purely deployed protocols in mainnet.

project-evaluations/src/data/evaluations/merces.json
```diff
@@ -1,10 +1,558 @@
...
+        },
+        {
+          "cited_text": "the latency experienced by a user is comparably high (e.g., 8.4 s as depicted in the screenshot). This discrepancy arises from the block time of the underlying blockchain.",
+          "source": "https://eprint.iacr.org/2026/850.pdf"
+        }
+      ]
+    },
+    {
+      "name": "Number of secrets",
+      "value": "2",
```
NicoSerranoP:
```suggestion
      "value": "1",
```

project-evaluations/src/data/evaluations/merces.json
```diff
@@ -1,10 +1,558 @@
...
+      "notes": "The user holds two secrets: their host-chain wallet key, used for the public ERC-20 deposit and withdraw legs, and a separate sender secret key for the private account whose public key is registered at an index in the on-chain ID-registry. Spends prove knowledge of the sender key, not the wallet. Per-transfer ECDH keypairs that encrypt the secret shares for the MPC nodes are sampled fresh and never persisted.",
```
NicoSerranoP:
```suggestion
      "notes": "The user only needs his Ethereum wallet to use any of the Merces operations: use his normal public key pair to deposit and withdraw ERC-20 tokens into the Merces contract. For internal private transfers, the interfaces ask users to sign a message that will be used to generate a new keypair to store in the on-chain ID-registry. Per-transfer ECDH keypairs that encrypt the secret shares for the MPC nodes are sampled fresh and never persisted.",
```

project-evaluations/src/data/evaluations/merces.json
```diff
@@ -1,10 +1,558 @@
...
+        {
+          "cited_text": "the latency experienced by a user is comparably high (e.g., 8.4 s as depicted in the screenshot). This discrepancy arises from the block time of the underlying blockchain.",
+          "source": "https://eprint.iacr.org/2026/850.pdf"
+        }
+      ]
+    },
+    {
+      "name": "Withdraw time",
+      "value": "0",
+      "notes": "Withdrawing is an ordinary on-chain action with no protocol-imposed waiting period. The user registers a withdraw intent in the action queue, and after the MPC network verifies it the contract releases an ERC-20 transfer to the receiver. The round-trip — register, MPC-process, verify, release — runs about 8.4 seconds end-to-end in the demo.",
```
NicoSerranoP, line 137: We can use that 8.4 seconds for the time-to-finality value. The problem is that I could not find any mention of "8.4" or "seconds" in the documentation nor the Merces blog post

project-evaluations/src/data/evaluations/merces.json
```diff
@@ -1,10 +1,558 @@
...
+        {
+          "cited_text": "Whenever a match is found, the corresponding shares of the transaction details are opened (i.e., revealed) to the regulator.",
+          "source": "https://core.taceo.io/articles/merces-deep-dive/"
+        }
+      ]
+    },
+    {
+      "name": "Enforcement entities",
+      "value": "Admin",
+      "notes": "The protocol operator controls enforcement. TACEO operates the MPC network that performs disclosure, and the administrator maintains the wallet registry and the blocklist that gate transfers. Per-transaction screening is delegated to Predicate, but the administrator sets policy and the Merces contract enforces it, with a signed warrant required for disclosure.",
```
NicoSerranoP:
```suggestion
      "notes": "The protocol operator controls enforcement. TACEO operates the MPC network that performs disclosure, and the administrator maintains the wallet registry and the blocklist that gate transfers. Per-transaction screening is delegated to Predicate, but the administrator sets policy and the Merces contract enforces it. A signed warrant is required to post a disclosure intent that the MPC network would execute.",
```

project-evaluations/src/data/evaluations/merces.json
```diff
@@ -1,10 +1,558 @@
...
+        },
+        {
+          "cited_text": "Whenever a match is found, the corresponding shares of the transaction details are opened (i.e., revealed) to the regulator.",
+          "source": "https://core.taceo.io/articles/merces-deep-dive/"
+        }
+      ]
+    },
+    {
+      "name": "Cryptographic verifiability",
+      "value": "No",
```
NicoSerranoP, line 415:
```suggestion
      "value": "Yes",
```

project-evaluations/src/data/evaluations/merces.json
```diff
@@ -1,10 +1,558 @@
...
+      "notes": "Each state transition is verified on-chain by a CoSNARK against the prior and new Merkle roots, so a corrupt MPC network cannot forge balances or steal funds. But the protocol rests on the MPC network under an honest-threshold assumption: the current deployment needs a majority of its operators to be honest.",
```
NicoSerranoP, line 416:
```suggestion
      "notes": "Each state transition is verified on-chain by a CoSNARK against the prior and new Merkle roots, so a corrupt MPC network cannot forge balances or steal funds. Each MPC node generates a CoSNARK to succesfully submit the updated state. The honest-threshold assumption implies that MPC nodes could collude between each other and decrypt the shared private state. The colluding nodes would not be able to arbitrarily modify any state because they would need valid CoSNARK proofs.",
```

project-evaluations/src/data/evaluations/merces.json
```diff
@@ -1,10 +1,558 @@
...
+        {
+          "cited_text": "We deduct the balance of the sender and add them to the receiver",
+          "source": "https://core.taceo.io/articles/merces-deep-dive/"
+        }
+      ]
+    },
+    {
+      "name": "Private Data Storage",
+      "value": ["Off-chain storage with on-chain commitment"],
+      "notes": "The contract stores only a commitment to the encrypted balance tree, keeping the current Merkle root and running an action queue for user intents. The authoritative plaintext lives off-chain with the MPC nodes: each party holds secret shares of the full balance Merkle tree, and transaction history is likewise retained off-chain in secret-shared form. On-chain artefacts are limited to commitments, nullifiers, ciphertexts for the nodes, Merkle roots, and proofs registered through the queue.",
```
NicoSerranoP, line 507:
```suggestion
      "notes": "The contract stores only a commitment to the encrypted balance tree, keeping the current Merkle root and running an action queue for user intents. The authoritative plaintext lives off-chain with the MPC nodes: each party holds secret shares of the full balance Merkle tree, and transaction history is likewise retained off-chain in secret-shared form. On-chain artefacts are limited to commitments, ciphertexts for the nodes, Merkle roots, and proofs registered through the queue.",
```

NicoSerranoP (approved):
Approved but there are some comments and change suggestions worth considering. I feel some properties values were not right

project-evaluations/src/data/evaluations/blanksquare.json
JohnGuilding: cheers

project-evaluations/src/data/evaluations/merces.json
JohnGuilding: > I think we should put the minimum external dependency time it takes between registering the intent and the external dependency performing the computation.

I'm not sure I'm on the same page with this comment. Finality would still be the chain it is deployed on? TACEO would need finality of it's own which I'm not sure exists.

> The problem with this value is that documentation does not state the specific time to finality in most of the cases

This may be common across coprocessors. Finality is a blockchain consensus concept, and to my knowledge is not a property of coprocessors. I think we're muddying terms by suggesting that a coprocessor has it's own finality. Let me know if this is not true according to your understanding.

project-evaluations/src/data/evaluations/merces.json
JohnGuilding: > We can use that 8.4 seconds for the time-to-finality value

The underlying chain finality would be longer since it is based on Ethereum's 2 epochs?

project-evaluations/src/data/evaluations/merces.json
JohnGuilding: There are no nullifiers?

project-evaluations/src/data/evaluations/merces.json
JohnGuilding: makes sense thanks

project-evaluations/src/data/evaluations/merces.json
JohnGuilding: checking this, you have source? From using it yourself? Actually I think I tried demo and didn't store an extra secret

project-evaluations/src/data/evaluations/merces.json
JohnGuilding: This is a good point, I'm thinking about this. I changed how we evaluated this property in this PR you already reviewed https://github.com/privacy-ethereum/private-transfers-benchmarks/pull/286.

project-evaluations/src/data/evaluations/merces.json
NicoSerranoP: Good point. In my perspective for private transfers was "how long it takes for my private transfer to be executed and unchangeable (finality)". For example here:

0. I have 5 ETH (time = 0)
1. I created an intent to "deposit 5 ETH to Merces" (time = 0)
2. I publish my intent so the coprocessor/MPC network can execute it (time=12 if Merces is in mainnet)
3. Coprocessors execute my intent and register 5 ETH to me inside Merces (time = 12 + N)

Finality time = 12 + N


I will resolve this conversation because I do not feel it is worth it to go deeper and change the finality definition that we currently have (which is that finality depends on the chain it is deployed). I can create an issue but it is not a priority :)

Issue: https://github.com/privacy-ethereum/private-transfers-benchmarks/issues/319

project-evaluations/src/data/evaluations/merces.json
NicoSerranoP: Yes you are right I was thinking something like "Finality": "Underlying chain + 8.4 seconds"

https://github.com/privacy-ethereum/private-transfers-benchmarks/issues/319

I will resolve this because we can discuss it in the issue is needed 😊

project-evaluations/src/data/evaluations/merces.json
NicoSerranoP: It was from using the demo at https://merces.taceo.io/:

1. connect wallet
2. sign message
3. ready to use (I assume the signed message is the seed for the internal ID-registry keypair. Otherwise they will ask you to back up some seed phrase or private key)

project-evaluations/src/data/evaluations/merces.json
NicoSerranoP: No because it is a account based model where each transfer updates the user's encrypted balance. Similar to HE but MPC implies that each network node contains their partial encrypted share. If they all collude they will get the user real balance.

The TACEO docs show results when searching for Nullifiers but these are mostly for distributed identity solutions (e.g. World) and not Merces itself

JohnGuilding: It seems to be related to merces as well, cm+f for "nullifier" in https://core.taceo.io/articles/merces-deep-dive/ has 12 hits and is passed in as function argument in core functions.

project-evaluations/src/data/evaluations/merces.json
JohnGuilding: Changing this in verifiability PR

project-evaluations/src/data/evaluations/merces.json
JohnGuilding: changing to 1

Issue #202, opened 2026-05-19
Consider basing evaluation on non-default behaviour


JohnGuilding:
## Description

We are currently evaluating issues based on the default behaviour of the protocol.

We could base it on what is possible even if it is prohibitive for the user:
https://github.com/privacy-ethereum/private-transfers-benchmarks/pull/168#discussion_r3262133218

## Acceptance Criteria

- Decide whether to keep things the same or change them
- Change rules
- Changes evaluations for all protocols

NicoSerranoP:
After looking at the third-party inspectability property of zERC20 I think it is good to evaluate projects based on their defaults. Users will find it hard to setup their own indexer to listening to receiving txs. The issue is that most protocols (note-based and stealth addresses) have default indexers. I am not 100% sure how do we handle that. 

We should discuss it in a sprint refinement meeting

PR #284, opened 2026-06-18
feat(project-evaluations): default behaviour rule change

JohnGuilding:
# What this PR does

Partly resolves https://github.com/privacy-ethereum/private-transfers-benchmarks/issues/202

Changes rule to base evaluations on non-default behaviour

**Changing historical evaluations is out of scope of this PR and will be done in a future change.** So the issue will not be resolved until that is done. This is because we want to get some new evaluations in for the report asap, so this partial change will speed up getting those in

~~merge https://github.com/privacy-ethereum/private-transfers-benchmarks/pull/282 first~~

now, merge https://github.com/privacy-ethereum/private-transfers-benchmarks/pull/315 first

## Confirmation

> [!IMPORTANT]


project-evaluations/scripts/research-prompts.ts
```diff
@@ -64,6 +64,7 @@ VALUE FORMAT RULES:
...
+5. Grade on what the protocol makes POSSIBLE, not only its default configuration. When a determined user can reach stronger privacy by self-hosting infrastructure (their own RPC node, relayer, sequencer, or indexer), the value reflects that achievable best case even if the path is prohibitive for a typical user and the default is more exposed (e.g. Third-party inspectability = No when a user can run their own indexer or relayer to avoid observation). State the configuration the value assumes in the notes, and flag the more-exposed default as a caveat.
```
NicoSerranoP, line 67: Not sure if we should aim for non-default behaviour. Maybe for third-party inspectability we could add a new property value "No when user deploys X" that way we can separate from default "No" and possible-if-user-does-extra-step "No"

JohnGuilding: I like your original suggestion of reviewing based on non-default flow. So long as the non-default flow was running open source infra already built.

Few options:
1. We could change this rule to just be about third party inspectability
2. We leave as is
3. We add new property as you suggested above

I'm leaning towards option 1 or 2. Reason being is I think the bar for changing properties such be reasonably high at this point. E.g. if we are being inaccurate. 

Btw this is the PR that closes this issue https://github.com/privacy-ethereum/private-transfers-benchmarks/pull/314

NicoSerranoP: I will go with 2. We leave as is 😊 if you agree. Otherwise open to go with 1. We could change this rule to just be about third party inspectability if you feel it might add value to the dashboard.

Up to you 🫡

JohnGuilding: leaving as is

PR #285, opened 2026-06-18
fix(subgraph): worm withdraw with event handlers

NicoSerranoP:
# What this PR does

- Get gasUsed for WORM withdraws
- Fix wrong return in Veil Cash util function

## Confirmation

> [!IMPORTANT]

- [ ] I have linked a github issue to the PR if one exists.


NicoSerranoP:
@codex

subgraph/src/worm/utils.ts
```diff
@@ -0,0 +1,26 @@
...
+    if (log.topics.length === NUMBER_OF_TOPICS_IN_WITHDRAW_EVENT) {
+      if (log.topics[0].equals(WITHDRAW_TOPIC)) {
+        const emitter = log.address;
+        const source = Address.fromBytes(Bytes.fromUint8Array(log.topics[1].slice(12)));
+
+        if (emitter.equals(WETH_MAINNET_CONTRACT) && source.equals(BETH_TO_ETH_CONTRACT)) {
+          return log;
+        }
+      }
+    }
```
0xmad: Three nested ifs is not a good pattern

0xmad (commented):
@NicoSerranoP thanks, just one comment

subgraph/src/worm/utils.ts
NicoSerranoP: You are right, I do not like that code either. I am not 100% sure how to approach it:

1. if we skip the first "topics length" if-block then the one below could break if there are no topics.
2. if we skip the second "first topic is withdraw topic" if-block then the one below could break because topics[1] could be anything and not an eth address
3. if we skip the third "emitter is WETH and the source is BETH_TO_ETH" if-block then we could consider other txs that have withdrawal events from different parameters


I tried putting different scenarios:

- 1 and 2 together in one if. It throws an error when topics do not exist
- 2 and 3 together in one if. It throws an error when there is a Withdrawal event but the second parameter was not an address

0xmad:
```
const isWithdrawTopicLength = log.topics.length === NUMBER_OF_TOPICS_IN_WITHDRAW_EVENT;
const isWithdrawTopic = log.topics[0].equals(WITHDRAW_TOPIC);
const isWethMainnetEmitter = emitter.equals(WETH_MAINNET_CONTRACT);
const isBethSource = source.equals(BETH_TO_ETH_CONTRACT);
  
if (isWithdrawTopicLength && isWithdrawTopic && isWethMainnetEmitter && isBethSource) {
  return log
}
```

0xmad (approved):
@NicoSerranoP thanks!

subgraph/src/worm/utils.ts
NicoSerranoP: Thanks! I changed a little bit so we that we would parse the source variable only with topic length greater than one. The code is:

```javascript
    const isCorrectTopicLength = log.topics.length === NUMBER_OF_TOPICS_IN_WITHDRAW_EVENT;
    const isWithdrawTopic = log.topics[0].equals(WITHDRAW_TOPIC);
    const isWETHMainnetEmitter = emitter.equals(WETH_MAINNET_CONTRACT);

    if (isCorrectTopicLength && isWithdrawTopic && isWETHMainnetEmitter) {
      const source = Address.fromBytes(Bytes.fromUint8Array(log.topics[1].slice(12)));

      if (source.equals(BETH_TO_ETH_CONTRACT)) {
        return log;
      }
    }
```

I just deployed a subgraph and it synced 100%

PR #310, opened 2026-06-24
Feat/platus evaluation

JohnGuilding:
Platus protocol evaluation plus operation-path blocklist sweep rule. Stacked on nightfall.


NicoSerranoP (approved):
I modified multiple notes. This evaluation was harder for the AI agent, and after reading the docs I would say that these were AI written as well (too technical and structured in a specific way): 

1. It looks like the agent was reading a non-existing page for compliance: https://docs.platus.xyz/compliance . I believe it is an hallucination because I tried to find it in the Wayback Machine and could not find it: https://web.archive.org/web/20260000000000*/https://docs.platus.xyz/compliance

2. There were multiple mentions of a JoinSplit circuit that is not present in the documentation. I believe it was referring to the Action circuit, the main one that allows private interactions on the shielded pool.

It would be interesting to explore why the agent had these two confusions.

PR #314, opened 2026-06-24
Feat/possible-behaviour re-grade

JohnGuilding:
Re-grade Third-party inspectability and related properties on possible (self-hosted) behaviour, plus the meta-language writing rule. Stacked on plasma-fold.


NicoSerranoP (approved):
Let's go! 🚀

